PR #27, opened 2015-07-24
Changing image file names

timstephens24:


drzoidberg33:
I'm not going to merge this as I am rather going to clean up the images folder and add the real Kodi image.

PR #31, opened 2015-07-25
Minute typo

iAmMrinal0:
'consider' instead of 'conside'


drzoidberg33:
I'll fix this, but any future requests need to be made to the dev branch only.

PR #32, opened 2015-07-25
Bug when most active user is "Local"

iAmMrinal0:
When the most active user is a "Local" user i.e. user who uses Plex without signing in(accessing in local networks mostly DLNA), then a['rows'][0]['user_id'] is set to None. This fixes it.


iAmMrinal0:
Well, just noticed the master branch and you seem to have fixed it.

PR #37, opened 2015-07-29
Changed href=Config to href=Settings

Mirabis:
My plex server went down a lot with the 12.6 update and i kept clicking settings... and got redirected to wrong url :(


iAmMrinal0:
Can you remove the extra newline in the end, for consistency?

drzoidberg33:
Thanks for taking the time but I had to change everything over to Bootstrap 3 so fixed these while I was at it.

PR #57, opened 2015-08-12
Systemd service file

nikdoof:
A systemd service file for PlexPy, should work on most systemd platforms.


drzoidberg33:
Thanks for the help.

Please can you make pull requests to the dev branch only.

PR #66, opened 2015-08-15
Feature allow custom profile pic

### LINKED ISSUE #15: Feature Request: Option to override user's profile picture URL
Currently users have their profile pictures as a letter.  It would be nice to be able to specify a different URL for the location of the user profile instead of the Gravatar URL.

jroyal:
Should close #15. I saw that you had a todo to refactor the datafactory code into proper getter and setters. I piggybacked on the getter, and created my own setter. 

I actually have a lot of that refactored, but I wanted to push that in a separate pull request so that way you could decide if you wanted it or not.

Btw, what was the reasoning behind user_id vs username? I see it all over the place and I wasn't sure why it was needed.

### DIFF
--- a/data/interfaces/default/edit_user.html
+++ b/data/interfaces/default/edit_user.html
@@ -93,6 +93,7 @@ DOCUMENTATION :: END
                     if ($.trim(friendly_name) !== '') {
                         $(".set-username").html(friendly_name);
                     }
+                    $("#user-profile-thumb").attr('src', thumb);
                 }
             });
         % else:
@@ -106,6 +107,7 @@ DOCUMENTATION :: END
                     if ($.trim(friendly_name) !== '') {
                         $(".set-username").html(friendly_name);
                     }
+                    $("#user-profile-thumb").attr('src', thumb);
                 }
             });
         % endif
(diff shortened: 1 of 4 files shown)


drzoidberg33:
Thanks, looks good going to test now.

Ye, the datafactory code is a bit of a mess, in the beginning I just shoved everything in there which I couldn't justify giving it's own class. As I go along I'm trying to move those functions into their own classes if needed.

It's probably a good idea to move user-related functions to it's own class going forward, the entire user section is really in need of an overhaul.

Regarding the the user_id and username, the username is really just there as a fallback. Everything should work off a user_id but sometimes there isn't a user_id to work with (like in the case with the Local user). I want to remove this though and only work with user_id going forward, the Local user should be given the user_id of -1.

If you have ideas on how to handle this better I'd love to discuss them.

jroyal:
Is the only user that doesn't have a user_id the local user, or is it possible to have more than that? I don't have any in my table so I'm not sure what would cause that to happen.

drzoidberg33:
The only time this happens is with the Local user and if somebody imports a PlexWatch database where their own username is no longer consistent with the data in the PlexWatch db.

PR #69, opened 2015-08-15
Most watched movie is now on home page

jroyal:
I discovered that your thread on the forums are a way better place to look for something to work on. Most people don't bother actually creating github issues for things. I saw this in that thread somewhere.


drzoidberg33:
Ooh nice. Thanks.

PR #71, opened 2015-08-15
Recently added TV Shows display title of the show

### LINKED ISSUE #70: Request: Include show title in recently added list on home page
From the forums

> @wally007 said:
> can you please reconsider or add an option what details are displayed for recently added tv shows/movies ?
> If tv show doesnt have a poster ( any recorded tv, home movie or tv show which is not listed on thetvdb ) I have no idea what was added since plexPy displays only season number and not the show title.
> 
>

jroyal:
Should close #70. The only thing I would be slightly worried about is what happens when a show has a particularly long name, but for now this should satisfy that request.

image


drzoidberg33:
Could you just add the new parameter to the docs in the header of recently_added.html. I see parent_title was added there.

jroyal:
Done. I had been meaning to ask if those were generated somehow or not. Will make sure I do that from now on.

drzoidberg33:
Nope, hand typed... all of them :dizzy_face:

PR #79, opened 2015-08-17
Two digit season and episode numbers and FreeNAS init script

JonnyWong16:
Added two digit season and episode number parameters for notifications (ie. for S00E00 formatting).
There is probably a better way to name the {season_num00} and {episode_num00} parameters for notifications. Feel free to rename them.

Also, init script for FreeNAS.


drzoidberg33:
Thanks!

PR #95, opened 2015-08-18
Update to homepage statistics

JonnyWong16:
- Added most popular movie to homepage
- New setting to toggle statistics based on play duration instead of
  play count


drzoidberg33:
Thanks! I've already written a function to convert the duration to human readable - will commit this now.

PR #96, opened 2015-08-18
More stylized homepage statistics and other style changes

JonnyWong16:
- Fixed some formatting on user page and posters on info page
- Fixed tab spacing in css file
- Fallback to show poster if season poster is unavailable
- Dynamically adjust user recently watched items
- Style changes
- Added missing metadata
- Don't show metadata field if data is unavailable

### DIFF
--- a/data/interfaces/default/css/plexpy.css
+++ b/data/interfaces/default/css/plexpy.css
@@ -513,7 +513,7 @@ input[type="color"],
     left: 0;
     float: left;
     text-align: left;
-    padding: 2px 2px 6px 2px;
+    padding: 5px 8px;
     font-size: 12px;
     color: #eee;
     text-overflow: ellipsis;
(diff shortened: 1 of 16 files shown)


drzoidberg33:
Thanks for the styling fixes and improvements.

The home stats posters are broken now though, they're overflowing their containers.

JonnyWong16:
I made it look like that.
Home Posters

drzoidberg33:
Ah. It looks broken mainly because they're not the same size as the user and platform cards.

JonnyWong16:
I can't think of a way to change the user and platform cards since they don't have a poster. I'm open to ideas if you have some.

drzoidberg33:
I'd just make sure the cards stay the same size and the content should stay in the bounds of the cards.

Probably just best to leave the posters the same size as they originally were.

JonnyWong16:
Home page statistic cards have been reverted.

drzoidberg33:
Thanks, love the changes. You fixed a lot of stuff that annoyed me about the old styles that I never got round to fixing.

PR #116, opened 2015-08-23
Quick fix for high contrast mode

JonnyWong16:
Not sure if this is the best way to fix this, but it works for now


drzoidberg33:
I really don't think this is a maintainable way to go forward.

Either we must go back to using src tags for all images and a Python function to fix aspect ratio issues or use a single javascript function to do this. Having the HC checks all of the place is not a proper solution.

JonnyWong16:
I agree. This might not even be worth getting to work for the couple of people using high contrast mode.

Alternatively, the script can be put into a GreaseMonkey/TamperMonkey script that will do the image replacements in the user's browser.

drzoidberg33:
I like the idea of having the user fix this on their end with a script like you suggested. There are probably like 0.00001% of people using PlexPy with high contrast mode.

JonnyWong16:
Custom userscript available here:
https://github.com/JonnyWong16/plexpy-highcontrastmodefix

PR #118, opened 2015-08-23
Add group editing of users from the user table

JonnyWong16:
From the user table, you can now:
- Change the friendly name
- Toggle notifications
- Toggle keep history
- Purge all user history (as a group selection)

**This feature is still experimental, bugs are expected!**

Show/hide users is just a blank toggle for now. I think a column like 'hide_user' needs to be added to the user database table to enable something like this.


drzoidberg33:
Looks good. Only problem I've seen so far is the popover on the Edit Mode button, when you resize the window it stays in the same spot.

PR #125, opened 2015-08-23
Adds ip_addres to notification options

JonnyWong16:
This is iffy because the IP address is unknown when state == 'play', which is when it's the most useful.

### DIFF
--- a/data/interfaces/default/settings.html
+++ b/data/interfaces/default/settings.html
@@ -767,12 +767,16 @@ available_notification_agents = notifiers.available_notification_agents()
                         <table>
                             <tbody>
                             <tr>
+                                <td width="150"><strong>{server_name}</strong></td>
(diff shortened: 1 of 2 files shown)


drzoidberg33:
Errr, I don't like this. It should be all or nothing, I don't want half-baked features that just cause confusion, which is why I didn't include this option in the first place.

Not having the IP check run first ensures that notifications are fired off as quickly as possible without having to wait for the log reader to do it's thing which can take some time, especially if there are multiple streams firing at the same times.

IMO an IP address is not all that useful on it's own anyway, it's useful for geo-location purposes after the fact but I cannot think of a useful purpose to have it there otherwise. If it's to identify devices rather rename the device to something more informative.

If Plex implements this into their API one day then it can be added.

PR #126, opened 2015-08-23
History tables

JonnyWong16:
I think the confirm modal should me moved into it's own file (similar to the ip address and stream detail modals) since we are using in everywhere.

I'm not that good with the back-end python stuff, so I'm not sure how to implement it. But I can try, just copy and paste the code for the other modals?

### DIFF
--- a/data/interfaces/default/css/plexpy.css
+++ b/data/interfaces/default/css/plexpy.css
@@ -1682,6 +1682,11 @@ input[type="color"],
 .history-title .popover.right .popover-content {
     padding: 5px 8px;
 }
+.history-thumbnail {
+    background-position: center;
+    background-size: cover;
+    width: 80px;
+}
 .edit-user-toggles {
     padding-right: 10px;
 }
--- a/data/interfaces/default/history.html
+++ b/data/interfaces/default/history.html
@@ -14,7 +14,11 @@
             <span><i class="fa fa-history"></i> History</span>
         </div>
         <div class="button-bar">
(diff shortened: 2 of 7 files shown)


drzoidberg33:
This still needs a little work. Currently it doesn't remember your selected deletes when refreshing the table, like when changing pages, altering the sorting or changing the page length.

While there is nothing wrong with that in itself there is nothing telling the user this is going to happen. The user should either be prevented from changing the state of the initial page or it should remember the options across table state changes.

When I initially implemented the delete option I tried using the Datatables Select extension (https://datatables.net/extensions/select/examples/) however there is a bug somewhere in their code that causes all items on the page to be selected when the table state changes. I didn't want to waste too much time trying to track this down so I opted for a simpler solution. Might be worth it to explore that avenue again so you don't need to write a bunch of new code.

JonnyWong16:
Added a rowCallback function to show the remembered selected state.

I will also fix this in the users table when you verify this works.

drzoidberg33:
Thanks. Looks to be working now, good job. I've run into an issue twice where the pop-over on the Delete Mode button lands up in the middle of the screen - I just cannot remember what I did to replicate this.

Merging it and can hopefully track that down if somebody reports it.

drzoidberg33:
Ah, can reproduce it now. If you enable delete mode, then scroll to the bottom of the page so the pop-over is no longer visible, then change the window dimensions - it forces the pop-over down.

JonnyWong16:
Gotcha, I'll look into it.

Maybe just a bootstrap alert box instead of a popover.

drzoidberg33:
I don't really mind, I kind of like the way it is now - it's non-obtrusive but gets the info across. I trust your judgement on that, UI has improved a lot since you started helping out.

PR #127, opened 2015-08-24
Miscellaneous fixes

JonnyWong16:
Sorry for so many pull requests.


drzoidberg33:
It looks like you've reverted some fixes I made previously here for the posters. The bigger posters (i.e. not the home stats ones) should all be pulled at 300x450 resolution to satisfy high dpi screen setups. This mirrors what Plex Web uses.

JonnyWong16:
Image sizes reverted

PR #134, opened 2015-08-25
Change alert message and button colour

JonnyWong16:


drzoidberg33:
That works well, simpler.

PR #139, opened 2015-08-26
Create body container so scrollbar is beneath navbar

JonnyWong16:


JonnyWong16:
Note to self: fix error alert on settings page for top_list_count when merged.

PR #142, opened 2015-08-27
Dashboard activity

JonnyWong16:


drzoidberg33:
The bif index thumbs are not working anymore, they're being loaded but not being displayed. Think it may have been something getting messed up in the latest merges.

JonnyWong16:
I can't get my server (version 0.9.12.11) to generate BIF index thumbs. It's not even showing BIF thumbs on Plex/Web. I know I have them because I've seen them before.

PR #143, opened 2015-08-28
Miscellaneous fixes

### LINKED ISSUE #106: Add last 5 watched items on the Homepage
It would be good to have the last 5 or 10 watched items (configurable if possible) on the homepage with following details as a small table: user - date/hour - movie/show title 

Instead of going to the history page to find out recently watched items.

JonnyWong16:
Resolves #106.

### DIFF
--- a/data/interfaces/default/info.html
+++ b/data/interfaces/default/info.html
@@ -73,7 +73,7 @@ DOCUMENTATION :: END
                         <span><i class="fa fa-chevron-right"></i></span>
                         <span><a href="info?item_id=${data['parent_rating_key']}">Season ${data['parent_index']}</a></span>
                         <span><i class="fa fa-chevron-right"></i></span>
-                        <span><a href="#">${data['title']}</a></span>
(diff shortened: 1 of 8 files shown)


drzoidberg33:
The new home stat is missing data causing it to throw an error (platform_type missing).

Just make sure that all the home stats return the exact same fields, even if the field is not applicable for that specific stat (it should then just return a blank value) - this ensures things don't bug out when calling that variable in the mako template.

data/interfaces/default/info.html
```diff
@@ -73,7 +73,7 @@
                         <span><i class="fa fa-chevron-right"></i></span>
                         <span><a href="info?item_id=${data['parent_rating_key']}">Season ${data['parent_index']}</a></span>
                         <span><i class="fa fa-chevron-right"></i></span>
-                        <span><a href="#">${data['title']}</a></span>
+                        <span><a href="#">Eipsode ${data['index']} - ${data['title']}</a></span>
```
drzoidberg33: Typo on Episode.

PR #145, opened 2015-08-29
Initial implementation of server statistics

JonnyWong16:


drzoidberg33:
Thanks, looks good. Was thinking maybe the new section should be named "Library Stats". Server stats sounds like it has to do with server resources. Thoughts?

JonnyWong16:
Good point. "Library Statistics" makes more sense.

drzoidberg33:
@JonnyWong16 

I think you should also consider limiting the amount of sections pulled (or let the user set a limit) as some people have ridiculous amounts of different sections. Maybe also allow disabling it completely - for me I don't really find much value in knowing how many items are in my library so would be nice to just disable it.

JonnyWong16:
I'll look into it. There's still a lot that can be done with these statistics.

PR #163, opened 2015-09-06
Fix for 5 digit library stats

JonnyWong16:


drzoidberg33:
Cool additions. Thanks.

Issue #115, opened 2015-08-23
When Users are deleted from Plex, they are not deleted from PlexPy


zenjabba:
Deleted 50 users from Plex, those users not deleted from PlexPy, even after requesting "refresh users"

JonnyWong16:
The deletion of users should not be automatic with "Refresh users". Some people might want to keep data from users removed from Plex.

With that being said, implementing a separate "Refresh and delete users" function could be done.

drzoidberg33:
Exactly what JW said.

You can use the purge option in the user edit screen to remove history items (although the user names will still show in the user list).

JonnyWong16:
I think the purge option on the user edit screen is broken, but I just implemented group purging in #118. Just purging history, not deleting users.

drzoidberg33:
@JonnyWong16 What exactly is broken? I tested it quite a few times and worked. Tested it now again and works - so some more detail would be helpful.

JonnyWong16:
@drzoidberg33 It could just be me. The javascript confirm function doesn't work for me in Chrome. Maybe I have something blocking it? It works in other browsers.

I might change it to use the bootstrap modal confirm box, like the one I implemented on the users table. It's prettier. =P

drzoidberg33:
That's weird. I use Chrome primarily and haven't had issues. The bootstrap modal is a lot prettier, so go ahead :smile:

zenjabba:
I've tried to use this function, but the purge function doesn't do anything (tried in Safari and Chrome). The Logs are showing deletion taking place, when the "All Users" screen is still showing the users.

2015-08-30 23:21:25 INFO    PlexPy DataFactory :: Deleting all history for user id 4551217 from database.
2015-08-30 23:21:25 INFO    PlexPy DataFactory :: Deleting all history for user id 4544296 from database.
2015-08-30 23:21:25 INFO    PlexPy DataFactory :: Deleting all history for user id 4419732 from database.

![screenshot 2015-08-30 23 26 56](https://cloud.githubusercontent.com/assets/679864/9571644/a07c38e2-4f6e-11e5-9c7b-db6f6d57e889.png)

drzoidberg33:
@zenjabba As mentioned in this very issue, the purge option only removes history for the user, it does not delete the actual user's entry. According to the screenshot above it looks like it's working exactly how it is intended to work.

JonnyWong16:
Delete users added in v1.2.8

Issue #86, opened 2015-08-17
FR: Option to ignore a user from the overviews


joswart:
I am only interested in what my friends view on my system. I already have a pretty good feeling of what I am viewing myself. It would be nice if one could selectively remove accounts from the display, so one could just see how many resources those others are claiming.

drzoidberg33:
Do you still want to keep history for the user but just not display the data?

joswart:
Yeah, would like to be able to dingle out groups of users. 

Jord Swart
Engerstraße 21a
D-40235 Düsseldorf

e: jord.swart@jordswart.org
t: +49 211 - 468 768 85
m: +49 162 - 214 74 85

drzoidberg33:
Dingle out? :laughing: I hope not.

Okay, so it must continue to log history for these users but you should be able to toggle whether the user shows up in any report.

There is already a per-user option to disable logging completely but then obviously you cannot re-enable it later and see old history for that user.

joswart:
I meant single out, but I guess you understood ;-)

Yes, basically the way you describe it, show me the history for group a and then show me the history for group b.

I wasn’t aware there is an option to disable certain users, where is it hidden? Or is there a fine manual somewhere I haven’t read?

Thanks for the quick responses!

drzoidberg33:
If you go to the Users page and select the user you want, on their profile page there is an edit icon (the pencil) next to their username. This will bring up options for that user.

PR #167, opened 2015-09-08
Hide users

JonnyWong16:
I'm not sure if applying the user filter to dashboard activity or synced items is a good idea.

Also, maybe default the "Local" user as hidden (show_user = 0)?

Reference #115, #86.


JonnyWong16:
What needs to be done to fix this up?

JonnyWong16:
Closing.

Delete user in 1c00f82 instead.

Issue #77, opened 2015-08-16
Allow grouping of played items


vanstinator:
Allow the grouping of played items so a user sees

10:00 User 1 Plays item 1

Instead of 

10:00 User 1 Plays item 1
10:10 User 1 Plays item 1
10:45 User 1 Plays item 1

There are some caveats. Imagine for instance

10:00 User 1 Plays item 1 at quality x
10:05 User 1 Stops item 1
10:10 User 1 Plays item 1 at quality z
10:15 User 1 Stops item 1
10:30 User 1 Plays item 1 at quality y
10:36 User 1 Stops item 1
10:40 User 1 Plays item 2
10:44 User 1 Stops item 2
11:00 User 1 Plays item 1 at quality y
11:30 User 1 Stops item 1
- How do we determine which items should be grouped? Does item 1 at 11:00 get grouped with the rest of item 1 before the user played item 2? How long do we wait before items are no longer groupable?
- The start and stop times vs duration. The start time of grouped item 1 is 10:00, the stop time is 11:30 (if we count in that last one). That's 1:30 of time passed but only 36 minutes of the item has been played. This should be correctly accounted for.
- The user has played the same item but at 3 different quality settings, should these items be grouped or not? If they are grouped we need to account for all three qualities somehow.
- Do we leave items separate if they're played from different devices? If User 1 started playing item 1 on device X, stops it and continues playing it on device Y that cannot be grouped.
- Do we write a new database table for this (like with PlexWatch) or do we query on the fly (not a simple task as there are many queries).
  Not impossible but it definitely is quite a challenge. PlexWatch trips up with a few of the above.

drzoidberg33:
Does this apply for all places where play counts get registered (e.g. graphs) or only in the tables?

vanstinator:
Graphs too.

"As a user I want the option to display "Plays" as a holistic viewing of a movie. If I stop and start a movie within a specified time frame I would like the data to show that I watched the movie one time, not 2-4 times depending on how many times I stop the film."

If that's our use case I think all data should be displayed with the assumption plays are grouped or not. In that case a new database table definitely makes the most sense so that we can use the same queries and just hit a table, which table being defined by a user option for grouping.

kylerw:
+1

ooii:
What about implementing it the same way as plexwatch (https://github.com/ljunkie/plexWatch#advanced-options---configpl)?

drzoidberg33:
[...]

PR #190, opened 2015-09-16
Group history

JonnyWong16:
Reference #77.

Need some major feedback on this, especially with the database query.

The query works, but the "from_table" is messy. Is there a way to query the "from_table" first, then feed the results back in to query the actual data? Unfortunately the data_tables.ssp_query() function doesn't work this way. :confused: 

Maybe just use that to create a new (temporary) table?


drzoidberg33:
Ye, this seems to work sometimes other times not and just makes it really hard to debug queries.

A new db table is the only solution I think that would make this viable.

I'm guessing we would only need one new table `session_history_aggregated` which is a clone of `session_history` and just contains two extra columns `children_ids` which contains a comma separated list of id keys which form part of the group and `reference_id` which is the id of the first item written (to make referencing the other tables easier).

The logic to group should then be done in a python function on a history write action, after the normal session_history write is completed it should then do the check for similar items to group and append the id from the `session_history` into the `children_ids` field if a suitable item is found otherwise write a new record to `session_history_aggregated`. 

This logic is basically what hasn't be completely fleshed out yet. One thing that works for one user may not work for another so it should probably err more on grouping a bit conservatively. Basically something like:

`if rating_key = rating_key and user_id = user_id and view_offset > view_offset then group`

Calling grouped history items is then as simple as changing the reference to the table name and id for the joins.

We would also need to create a script that would automatically generate this table from the existing data when a user upgrades.

JonnyWong16:
Even better, no new table needed. :smile: 

Just added `reference_id` to the `session_history` table. The query will either group by `id` or `reference_id` depending on whether we want the un-grouped or grouped history respectively. Then a `custom_where` is used on the `reference_id` to get the individual items from that group.

Everything seems to work for me so far.

The logic to group is only `rating_key = rating_key and user_id = user_id` (of the previous row). I didn't use `view_offset > view_offset` since I think the items should be grouped even if the user jumps back in time or plays the same item again. If a user plays the same item over and over again, I think it is reasonable to group the sessions as to not fill up the history.

Let me know what you think.

I also haven't looked at updating the code to include `reference_id` when importing a plexWatch database.

drzoidberg33:
I'm actually starting to like the solution you have here. I also see the settings option is descriptive on what exactly this does - it's not a true watch count grouping but just an option to hide pesky duplicates in the tables. I wasn't really on the same page until I read the description in the settings :smile:

While this doesn't really cover the main concern with https://github.com/drzoidberg33/plexpy/issues/77 it does cover one of two issues there imo. Issue one being the table duplicates, and issue two being a way to see a true watched count.

However a true watch count grouping really wouldn't make much sense where a time scale is present as if somebody watches half a movie on Monday then the other half on Thursday that is one "watch" technically but how do you plot that on a table with a time column or graph?

I think this would probably be something more suited to where total plays are displayed as a single value like in the stats cards (however we still work on time scales there) or in some new kind of stats in the user profiles pages.

Can I leave you with this one though:

screenshot 2015-09-16 22 48 10

While I don't think the play at 11:41 should be grouped, it would seem like a good idea to try and group the 18:30 and 18:41 plays. Your thoughts?

JonnyWong16:
How about this: Instead of checking the previous row, check the user's previous row to match the rating key.

So in your example, if USER A watches half a movie on Monday, and Thursday, but nothing else in between, then it can be counted as "one watch". USER B watching something on Tuesday won't break this continuity.

However, if USER A watched a TV show on Wednesday, it breaks the continuity, so there would be "three watches".

With this logic, the entries in your screenshot won't be grouped because it's by the same user. On the other hand, if "Susie" watched Adventure Time (awesome show btw), then Jeffery's three plays of The Office would be grouped.

This logic would could be a "watch" count, which would be independent of "play" count. The homepage statistics could probably be switched to this "watch" instead of "play". (Although "58 watches" sounds like someone owns 52 watches...)

As for graphing the "watch" count, I haven't thought that far ahead yet. This only makes sense on the tables.

JonnyWong16:
Is this grouping logic ok?

drzoidberg33:
Sorry, haven't had time to test again. Testing today.

drzoidberg33:
Working pretty well. One thing I don't like though is the icon on the table.

It pushes the date out of line with the non-expandable items. As below:

screenshot 2015-09-27 18 31 46

So if you could put a blank space to pad the non-expandable rows and change the icon to a plus icon (`fa-plus-circle`) which is generally what's used to show when something can be expanded.

drzoidberg33:
Woops, all the rows are now expandable (even if they're single results) :smile:

JonnyWong16:
I think it's because of switching branches constantly, the reference_id in the db didn't update. It happened to me. I had to manually run the SQL command to update the column.

Should we add a check, if reference_id == 0 is found in the database, then update the column?

data/interfaces/default/js/tables/history_table.js
```diff
@@ -46,13 +46,18 @@ history_table_options = {
...
                 if (rowData['stopped'] === null) {
                     $(td).html('Currently watching...');
+                } else if (rowData['group_count'] > 1) {
+                    date = moment(cellData, "X").format(date_format);
+                    expand_history = '<span class="expand-history-tooltip" data-toggle="tooltip" title="Show Detailed History"><i class="fa fa-plus-circle fa-fw"></i></span>';
+                    $(td).html('<div><a href="#"><div style="float: left;">' + expand_history + '&nbsp;' + date + '</div></a></div>');
                 } else {
-                    $(td).html(moment(cellData,"X").format(date_format));
+                    date = moment(cellData, "X").format(date_format);
+                    $(td).html('<div><a href="#"><div style="float: left;"><i class="fa fa-fw"></i>&nbsp;' + date + '</div></a></div>');
```
drzoidberg33: There just shouldn't be an `href` here. That's what I mean by my previous comment.

JonnyWong16:
Oops. Copy + paste without checking.

plexpy/monitor.py
```diff
@@ -0,0 +1,438 @@
+﻿# This file is part of PlexPy.
```
drzoidberg33: This class/file no longer exists. It's been split up into `activity_pinger` and `activity_processor`.

JonnyWong16:
Websockets makes testing this so much quicker. Don't have to wait until the session writes to the history. :smile:

data/interfaces/default/js/tables/history_table.js
```diff
@@ -231,24 +239,58 @@ history_table_options = {
...
+            // if no grouped rows simply toggle the delete button
+            if ($.inArray(rowData['id'], history_to_delete) !== -1) {
+                $(row).find('button[data-id="' + rowData['id'] + '"]').toggleClass('btn-warning').toggleClass('btn-danger');
+            }
+        } else {
+            // if grouped rows
+            // toggle the parent button to danger
             $(row).find('button[data-id="' + rowData['id'] + '"]').toggleClass('btn-warning').toggleClass('btn-danger');
+            // check if any child rows are not selected
+            for (var i = rowData['reference_id']; i <= rowData['id']; i++) {
```
drzoidberg33: Not quite sure if this is the right way to go about this. If you have a bad `reference_id` (like I had while testing) this will select all keys regardless of whether they're in the group or not.

For example I had grouped items with `reference_id` of 0 and it then proceeded to want to delete all items from id 0.

I unfortunately don't have any suggestions yet.

JonnyWong16: Update the database to make sure there are no `reference_id = 0`?

drzoidberg33: I haven't looked at it much detail yet but what if the ids for that group aren't sequential? 

What if user 1 watched grouped item id 1, 2, 3 and 5 and user 2 watched item id 4. The way I'm looking at it, it looks like it will just delete items 1 - 5.

JonnyWong16: I think I have an idea. I can use `GROUP_CONCAT` in SQLite to get a comma delimited string of individual row `id`'s associated with the `referencd_id`, then use that list instead of the for loop.

drzoidberg33: Muuuch better!

PR #195, opened 2015-09-19
Selective hiding home homepage statistics

JonnyWong16:
~~All cards are removed on first run after this update and the user is prompted to enable cards in the settings. I can change it to default as show all cards until the user disables them.~~

Default to show all cards until the user disables them.

Reference #106.


drzoidberg33:
Vey nice.

Just one bug I picked up, I doesn't appear I can switch between duration and count anymore for the home stats - the settings option has no effect and always shows duration.

drzoidberg33:
I'm going to merge this so long as I need to make some changes to the same file and don't want to have to merge things manually.

Please just check out that bug first thing when you get this. Thanks!

drzoidberg33:
Don't worry about the bug, I found the issue and was a very simple fix.

PR #198, opened 2015-09-20
Open in new tab

Hellowlol:


drzoidberg33:
Hey, thanks for the contribution but could you just resubmit it to the dev branch. Nothing is ever merged directly to master except when doing releases.

Hellowlol:
Ok. Why didn't you go for ws4py was a websocket library? Just wondering as it seems to play nice with cherrypy.

drzoidberg33:
Thanks for the suggestion, I'll take a look. I just had a look at what had been done with the Trakt.tv plugin for PMS and used that a base to get things started. I'm focusing now on re-writing the monitoring to accommodate the websocket events but would be really easy to drop in a different websocket library if I needed to.

Hellowlol:
I don't know if you have plans for the api or if it's just left overs from headphones, but if you are planning to use it (some of the calls in use atm seems like fetch data x from db, dump it as json) I would advice to use the api with a internal apikey so we only have one code base to maintain. If that is something you would be interested in I can help out.

drzoidberg33:
I most definitely want to provide some api endpoints but it's always at the bottom of the list of things to do. A lot of the stuff could probably be moved over from the webserve code to the api without much effort.

Hellowlol:
I have started on revamping the api, its going slow but ill move as much as possible to the api. The api will break tho, (dunno if anybody is using it) as i dont like the responses it gives atm. 

The new response will be: 

```
{"result":"success", "data": "somejson", "message": ""}
```

PR #200, opened 2015-09-22
initial api changes

Hellowlol:
This is just a initial pr so i can check with you if this is something you want or not?

### DIFF
--- a/PlexPy.py
+++ b/PlexPy.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
 # This file is part of PlexPy.
 #
 #  PlexPy is free software: you can redistribute it and/or modify
--- a/plexpy/api.py
+++ b/plexpy/api.py
@@ -29,7 +29,7 @@ import xmltodict
 
 cmd_list = ['getLogs', 'getVersion', 'checkGithub', 'shutdown',
             'getSettings', 'restart', 'update', 'getApikey', 'getHistory',
-            'getMetadata', 'getUserips', 'getPlayby']
+            'getMetadata', 'getUserips', 'getPlayby, getSync']
 
 
 class Api(object):
@@ -456,7 +456,7 @@ class Api(object):
         else:
             self.msg = 'Failed to find users ips'
 
-    def _getPlayby(self, time_range=30, y_axis='plays', playtype='total_plays_per_month', **kwargs):
+    def _getPlayby(self, time_range='30', y_axis='plays', playtype='total_plays_per_month', **kwargs):
 
         graph = graphs.Graphs()
         if playtype == 'total_plays_per_month':
--- a/plexpy/api.py
+++ b/plexpy/api.py
@@ -29,7 +29,7 @@ import xmltodict
 
 cmd_list = ['getLogs', 'getVersion', 'checkGithub', 'shutdown',
             'getSettings', 'restart', 'update', 'getApikey', 'getHistory',
-            'getMetadata', 'getUserips', 'getPlayby, getSync']
+            'getMetadata', 'getUserips', 'getPlayby', 'getSync']
 
 
 class Api(object):
--- a/plexpy/api.py
+++ b/plexpy/api.py
@@ -1,3 +1,6 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
 #  This file is part of PlexPy.
 #
 #  PlexPy is free software: you can redistribute it and/or modify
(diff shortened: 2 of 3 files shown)


drzoidberg33:
Thanks, looks good so far - I'll test later when I get home this evening.

The getHistory method used in here is deprecated though so you can remove it completely.

PlexPy.py
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+#!/usr/bin/env python
```
drzoidberg33: Any reason this is getting duplicated?

plexpy/api.py
```diff
@@ -13,82 +16,108 @@
...
-
         self.kwargs = None
-
         self.data = None
-
+        self.msg = None
         self.callback = None
+        self.result_type = 'error'
 
     def checkParams(self, *args, **kwargs):
```
drzoidberg33, line 51: The new changes break this function - the initial call from webserve.py expects this function to return something. It now no longer returns anything and none of these errors are going to get caught properly (e.g. you can use any apikey).

PlexPy.py
Hellowlol: Not at all, its just my editor inserting it, it will be cleaned up before i send a pr.

plexpy/api.py
Hellowlol, line 51: Thanks! I didnt check for that, only missing commands, invalid command etc.

Hellowlol:
I think its ready now, can you test the pr?

plexpy/api.py
```diff
@@ -13,86 +16,134 @@
...
+        self.out_type = out
+        self.result_type = 'error'
 
     def checkParams(self, *args, **kwargs):
 
         if not plexpy.CONFIG.API_ENABLED:
-            self.data = 'API not enabled'
-            return
+            self.msg = 'API not enabled'
         if not plexpy.CONFIG.API_KEY:
```
drzoidberg33: All the following if statements should either be `elif`s or this function should return something on each evaluation otherwise these `if`s are just ignored.

plexpy/api.py
```diff
@@ -13,86 +16,134 @@
...
+            try:
+                out = xmltodict.unparse(d, pretty=True)
+            except ValueError as e:
+                logger.error('Failed to parse xml result')
+                try:
+                    out = xmltodict.unpase({"response": {"result": "error", "message": str(e), "data": None}}, pretty=True)
+                except Exception as e:
+                    logger.error('Failed to parse xml result error message')
+
+        return out
```
drzoidberg33, line 129: `out` hasn't been initialised anywhere, going to throw an exception if none of the above conditions are true (or fail).

plexpy/api.py
Hellowlol: They simply adds a message. I want the function to run regardless so the response is made correctly.

drzoidberg33: What I mean is that that top argument is never going to be met, it's always going to be overridden by the last true `if` statement and therefor return the incorrect error message.

drzoidberg33:
Thanks! Commented on a few things but will have to test more when I get home later.

plexpy/api.py
```diff
@@ -13,86 +16,134 @@
...
+                logger.info(u"API :: " + traceback.format_exc())
+                out = '{"response": {"result": "error", "data": "None", "message": "error while composing output: "' + str(e) + '"}}'
+        if self.out_type == 'xml':
+            cherrypy.response.headers['Content-Type'] = 'application/xml'
+            try:
+                out = xmltodict.unparse(d, pretty=True)
+            except ValueError as e:
+                logger.error('Failed to parse xml result')
+                try:
+                    out = xmltodict.unpase({"response": {"result": "error", "message": str(e), "data": None}}, pretty=True)
```
Hellowlol: typo. should be unparse

plexpy/api.py
Hellowlol, line 129: Thanks.

drzoidberg33:
Just one more thing. When pulling the data from PMS/Plex.tv services directly, ask for it in `dict` format otherwise we get a case where the xml or json result is output as a string inside the result.

Using getMetadata as an example, just change the call to this:

`result = pms_connect.get_metadata(rating_key, 'dict')`

E.g. What we are getting:

screenshot 2015-09-27 18 04 41

What we want:

screenshot 2015-09-27 18 04 56

Hellowlol:
I would really like to remove the '@' from the attrs in the ordereddicts, but im but sure if i would mess up any other code .P

PlexPy.py
drzoidberg33: This duplicate line is still here.

Hellowlol: It slipped my mind, thanks for the reminder. I'll fix it when it get home from work.

Hellowlol:
rebased and squashed commits

plexpy/api.py
```diff
@@ -217,9 +289,211 @@ def _checkGithub(self, **kwargs):
...
+        history = user_data.get_user_unique_ips(kwargs=kwargs,
+                                                custom_where=custom_where)
+
+        if history:
+            self.data = history
+            return history
+        else:
+            self.msg = 'Failed to find users ips'
+
+    def _getPlayby(self, time_range=30, y_axis='plays', playtype='total_plays_per_month', **kwargs):
```
drzoidberg33: time_range must be a string (and I must fix bugs in the graph code).

plexpy/api.py
```diff
@@ -217,9 +289,211 @@ def _checkGithub(self, **kwargs):
...
+            start_date = kwargs.get('start_date', "")
+            custom_where = [['strftime("%Y-%m-%d", datetime(date, "unixepoch", "localtime"))', start_date]]
+
+        data_factory = datafactory.DataFactory()
+        history = data_factory.get_history(kwargs=kwargs, custom_where=custom_where)
+
+        self.data = history
+        return self.data
+
+    def _getSync(self, machine_id=None, user_id=None, **kwargs):
```
drzoidberg33, line 331: I see this is here but not in the commands list, still working on it?

Hellowlol, line 331: Nope, its done, but i didnt notice that one, must have missed it while i copy pasted

Hellowlol:
I think its good to go now :)

plexpy/api.py
```diff
@@ -13,86 +16,143 @@
...
+import cherrypy
+import re
+import hashlib
+import random
+import xmltodict
 
-cmd_list = ['getHistory', 'getLogs', 'getVersion', 'checkGithub', 'shutdown', 'restart', 'update']
+cmd_list = ['getLogs', 'getVersion', 'checkGithub', 'shutdown',
+            'getSettings', 'restart', 'update', 'getApikey', 'getHistory',
+            'getMetadata', 'getUserips', 'getPlayby, getSync']
```
majora2007: Shouldn't this be: 'getPlayby', 'getSync'? It looks like you are missing a comma.

Hellowlol: Its missing two '. Thanks!

drzoidberg33:
Thanks, looks like everything working now.

PR #205, opened 2015-09-24
Good day mate!

Hellowlol:


drzoidberg33:
This is great but not exactly what I had in mind with the dropdown I suggested.

It's now not possible to set a custom server on first run (e.g. if your server isn't local).

I was thinking more along the lines of something like this:

screenshot 2015-09-25 11 04 44

You should still have the ability to set a custom IP/hostname. Also not sure if you select a discovered server if it's necessary for the verify button as you're already discovered it is a working Plex server.

Hellowlol:
Ok, ill see if i cant add it the way you suggested. I totally forgot about remote servers.

drzoidberg33:
Another possibility is to move the Plex.tv authentication before the server selection in the wizard, then use the device discovery endpoint `https://plex.tv/api/resources?includeHttps=1`

Devices that return `provides="server" owned="1"` are servers that should be listed. If this doesn't work then fallback to local discovery. Just an idea.

Hellowlol:
Good idea! I didnt even know that was a possibility.

Hellowlol:
Can you point to the file and what line that the dropdown is on? I dont have access to plexpy from work.

drzoidberg33:
That screenshot is from Plex Web -> Settings -> Users -> Edit User -> Restrictions.

PlexPy doesn't have any implementation like this yet.

Hellowlol:
Oh, is the image just a mockup or have you used something like http://bootstrap-combobox-test.herokuapp.com ?

Arcanemagus:
It's a screenshot from plex.tv is I believe what he was saying.

Hellowlol:
Thanks, l2 read the responses

Hellowlol:
I added support for selectize (same as plex), The only thing im not happy about is that the size of the input field increases when its focused. I'm awefull at css, maybe you can figure it out?

JonnyWong16:
I tried testing this, it doesn't find the PMS on my network. Testing on windows, and PMS in a FreeNAS jail.

```
26-Sep-2015 16:41:49 - DEBUG   :: CP Server Thread-4 : Trying to find all pms on your network
26-Sep-2015 16:41:50 - DEBUG   :: CP Server Thread-4 : Didn't find any Plex Media Servers :(
26-Sep-2015 16:41:50 - DEBUG   :: CP Server Thread-4 : Found 0 servers
26-Sep-2015 16:41:50 - DEBUG   :: CP Server Thread-4 : []
```

Manually entering the IP address and Port number doesn't even let me verify. Nothing happens when I click the button.

Hellowlol:
And the port is open and gdm it not turned off?

JonnyWong16:
@Hellowlol Yes GDM is on.

I think the simplest solution is what @drzoidberg33 mentioned earlier. Authenticate plex.tv fist, and just use the API to get the list of servers.

Hellowlol:
Why is the plex token needed at all really?. :P, I dont like to have stuff online unless i have to.

JonnyWong16:
See https://forums.plex.tv/discussion/comment/1017346#Comment_1017346

Hellowlol:
Thanks for the link. It would be eady to add support for that. The "hard" part was to get all the data needed from  selectize.

JonnyWong16:
The function is already there `plextv.get_plextv_resources()`. It might help to look at how I populated the server libraries list in the Homepage Statistic settings. It should be similar for the servers list.

Issue #204, opened 2015-09-24
[Feature Request] Update database for moved library items


JonnyWong16:
Feature to update the PlexPy database when library items are moved (rating key is changed).
- If no metadata is found when generating info pages, use the Plex `/search` API to get a list of possible matches.
- Display the matches to the user to select the correct item.
- Update the PlexPy database with the new metadata, and generate the info page.

(This is mostly a note to myself to implement)

drzoidberg33:
Integrated.

PR #207, opened 2015-09-25
Update rating key for missing items

JonnyWong16:
I tested this by changing the various rating keys using SQLite DB Browser to simulate a "moved" library item, then visiting `info?item_id=rating_key`.

Reference #204.

### DIFF
--- a/data/interfaces/default/info.html
+++ b/data/interfaces/default/info.html
@@ -32,6 +32,14 @@ directors               Returns an array of directors.
 studio                  Returns the name of the studio.
 originally_available_at Returns the air date of the item.
 
+query :: Usable parameters
+
+== Global keys ==
+rating_key              Returns the unique identifier for the media item.
+media_type              Returns the type of media. Either 'movie', 'show', 'season', 'episode', 'artist', 'album', or 'track'.
+title                   Returns the title for the search query.
+
+
 DOCUMENTATION :: END
 </%doc>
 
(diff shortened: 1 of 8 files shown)


drzoidberg33:
This is pretty cool, will test more when I get home later.

drzoidberg33:
There are a few issues with this currently. For items without parents or grandparents (i.e. movies) this works great, however when you start changing parent and grandparent keys it doesn't do what the I think the user is expecting it to do (and also starts messing with the db consistency).

If I update something it should update all it's related keys. If a user updates an episode or a season it should automatically update everything for that entire show. I would create a map of the show/artist from `metadata/{key}/children` and then match by index and parentIndex to apply the new rating keys for all siblings.

I'm sure this would be doable, any pitfalls I'm not taking into consideration here?

JonnyWong16:
I just didn't want to make massive changes to the database, thus only changing one item at a time.

One thing. Tracks don't have a parentIndex, so they would need to be matched by title.

JonnyWong16:
Code now updates all the grandparent/parent/child rating keys for show/season/episode or artist/album/track.

You can test the rating key mappings with

```
/get_new_rating_keys?rating_key=<rating_key>&media_type=<media_type>
/get_old_rating_keys?rating_key=<rating_key>&media_type=<media_type>
/get_map_rating_keys?old_rating_key=<rating_key>&new_rating_key=<rating_key>&media_type=<media_type>
```

Note: the last one actually updates the database.

I will remove these functions when everything is verified to work properly. (Or leave them in, if you want)

Oh, and I added in the search feature. I figured since I used search to find matches for moved items, I might as well use that feature.

plexpy/datafactory.py
```diff
@@ -779,3 +779,190 @@ def delete_all_user_history(self, user_id=None):
...
+        # map rating keys pairs
+        mapping = {}
+        if old_key_list and new_key_list:
+            mapping = get_pairs(old_key_list, new_key_list)
+        
+        if mapping:
+            logger.info(u"PlexPy DataFactory :: Updating rating keys in the database.")
+            for old_key, new_key in mapping.iteritems():
+                # check rating_key (3 tables)
+                monitor_db.action('UPDATE session_history SET rating_key = ? WHERE rating_key = ?', [new_key, old_key])
```
drzoidberg33: Just a formatting thing here, could you just make sure that the lines don't get too long here. PEP8 default is 80 characters per line but I find that too short - I use 120 which is the default on PyCharm.

drzoidberg33:
Looks to be working well - I'm just testing some more.

data/interfaces/default/info.html
```diff
@@ -361,11 +374,122 @@ <h4 class="modal-title" id="myModalLabel">Confirm Delete</h4>
...
+                                    <p> with </p>
+                                    <p><span id="new_title"></span></p>
+                                    % if query['media_type'] != 'movie':
+                                    <p>All items for <strong>${query['grandparent_title']}</strong> will also be updated.</p>
+                                    % endif
+                                    <p>This is permanent and cannot be undone!</p>
+                                </div>
+                                <div class="modal-footer">
+                                    <button type="button" class="btn btn-dark" data-dismiss="modal">Cancel</button>
+                                    <button type="button" class="btn btn-danger btn-ok" data-dismiss="modal" id="confirm-update">Update</button>
```
drzoidberg33, line 485: Can we get this button to disable once clicked - not sure of what unexpected results may happen if you issue this request multiple times.

PR #214, opened 2015-09-28
Change to notify stopped only if less than watched percent

JonnyWong16:
- Also fix paused notifications sending at the end of items (right before stopping)

### DIFF
--- a/plexpy/activity_handler.py
+++ b/plexpy/activity_handler.py
@@ -16,7 +16,7 @@
 import time
 import plexpy
 
-from plexpy import logger, pmsconnect, activity_processor, threading, notification_handler
+from plexpy import logger, pmsconnect, activity_processor, threading, notification_handler, helpers
 
 
 class ActivityHandler(object):
@@ -66,8 +66,6 @@ class ActivityHandler(object):
 
     def on_stop(self, force_stop=False):
         if self.is_valid_session():
-            from plexpy import helpers
-
             logger.debug(u"PlexPy ActivityHandler :: Session %s has stopped." % str(self.get_session_key()))
 
             # Set the session last_paused timestamp
@@ -99,8 +97,6 @@ class ActivityHandler(object):
 
     def on_pause(self):
         if self.is_valid_session():
-            from plexpy import helpers
-
             logger.debug(u"PlexPy ActivityHandler :: Session %s has been paused." % str(self.get_session_key()))
 
             # Set the session last_paused timestamp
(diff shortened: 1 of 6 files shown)


plexpy/activity_handler.py
```diff
@@ -187,11 +194,11 @@ def process(self):
                                              view_offset=self.timeline['viewOffset'])
                     # Start our state checks
                     if this_state != last_state:
-                        if this_state == 'paused':
+                        if this_state == 'paused' and progress_percent < 99:
```
drzoidberg33: You shouldn't change it here, these `on_` functions do more than notifications. Especially the `stopped` state which triggers all the processing.

plexpy/activity_handler.py
```diff
@@ -83,8 +85,10 @@ def on_stop(self, force_stop=False):
             db_session = ap.get_session_by_key(session_key=self.get_session_key())
 
             # Fire off notifications
-            threading.Thread(target=notification_handler.notify,
-                             kwargs=dict(stream_data=db_session, notify_action='stop')).start()
+            progress_percent = helpers.get_percent(self.timeline['viewOffset'], db_session['duration'])
+            if progress_percent < plexpy.CONFIG.NOTIFY_WATCHED_PERCENT:
```
drzoidberg33: What if we're not using watched notifications and rely on stopped notifications or if we want both?

These notification changes should be something we can toggle - one setting toggle is fine with me. Something like `permit consecutive notifications` which you can disable to achieve this.

plexpy/activity_handler.py
```diff
@@ -66,6 +66,8 @@ def on_start(self):
 
     def on_stop(self, force_stop=False):
         if self.is_valid_session():
+            from plexpy import helpers
```
Hellowlol: Whats up with all the inline imports?

JonnyWong16: Copy and paste. Fixed.

drzoidberg33:
Regarding the notification stuff - I don't know if this is the right way to do it.

We are logging all the timestamps for notifications in the `notify_log` - I would write some getter/setters for these so we have easy access to retrieving them then use these values to determine whether notifications should fire off. All the logic should be done in the `notification_handler` - that code is a mess I know :smile:

JonnyWong16:
There's not much in `notify_log` to use...Time stamps of when a notification was sent aren't really the best for writing logic, besides maybe something like `on_stop > on_watched`.

What about just putting the logic right before "Build and send notification" in `notification_handler`? And still set the notification in `notify_log`.

PR #215, opened 2015-09-29
If we fail to parse a response, log the uri.

Hellowlol:


Hellowlol:
Would help to debug stuff like this: https://github.com/drzoidberg33/plexpy/issues/212

plexpy/http_handler.py
```diff
@@ -88,19 +88,24 @@ def make_request(self,
...
+                    else:
+                        output = request_content
+    
+                    if return_type:
+                        return output, content_type
+    
+                    return output
+                    
+                except Exception as e:
+                    logger.warn(u"Failed to parse %s to %s %s" % (uri, output_type, e))
```
drzoidberg33, line 107: You've got a typo here - there is no variable `output_type`.

Hellowlol, line 107: thanks!

PR #217, opened 2015-09-29
If we fail to parse a response, log the uri

Hellowlol:
and the original response

### DIFF
--- a/plexpy/http_handler.py
+++ b/plexpy/http_handler.py
@@ -1,3 +1,6 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
 # This file is part of PlexPy.
 #
 #  PlexPy is free software: you can redistribute it and/or modify
@@ -14,11 +17,10 @@
 #  along with PlexPy.  If not, see <http://www.gnu.org/licenses/>.
 
 from plexpy import logger, helpers
-
 from httplib import HTTPSConnection
 from httplib import HTTPConnection
-
 import ssl
+
 
 class HTTPHandler(object):
     """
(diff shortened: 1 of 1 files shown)


Hellowlol:
We try again :D

plexpy/http_handler.py
```diff
@@ -88,19 +88,25 @@ def make_request(self,
...
+                    else:
+                        output = request_content
+    
+                    if return_type:
+                        return output, content_type
+    
+                    return output
+                    
+                except Exception as e:
+                    logger.warn(u"Failed format response from uri %s to %s error %s. The original response was %s" % (uri, output_format, e, request_content))
```
drzoidberg33: Not so sure we should log `request_content` here. That may be a very large amount of text.

Hellowlol: Yes, but the its easy to see where is went wrong then. But its up to you. We should change the memory log too, its like having a memory leak.

But your the man, if you want it removed, let me know.

drzoidberg33: Ye, I don't want an entire xml dump being pushed into one line of the log. If you want to dump the xml then rather dump it to a separate file.

Hellowlol:
done

plexpy/http_handler.py
```diff
@@ -88,19 +90,25 @@ def make_request(self,
...
+
+                    if return_type:
+                        return output, content_type
+
+                    return output
 
-                if return_type:
-                    return output, content_type
+                except Exception as e:
+                    logger.warn(u"Failed format response from uri %s to %s error %s" % (uri, output_format))
```
drzoidberg33: You're missing an argument here. `e` is not being passed.

Hellowlol: lol, sorry, it got late yesterday.

PR #218, opened 2015-09-29
Change to notify stopped only if less than watched percent

JonnyWong16:
- Also fix paused notifications sending at the end of items (right
  before stopping)

Let's try this again. It only stops the notification right before it sends the notification (in `notification_handler`, and still writes the notification to `notify_log`.

### DIFF
--- a/plexpy/notification_handler.py
+++ b/plexpy/notification_handler.py
@@ -13,7 +13,7 @@
 #  You should have received a copy of the GNU General Public License
 #  along with PlexPy.  If not, see <http://www.gnu.org/licenses/>.
 
-from plexpy import logger, config, notifiers, database
+from plexpy import logger, config, notifiers, database, helpers
 
 import plexpy
 import time
@@ -33,6 +33,8 @@ def notify(stream_data=None, notify_action=None):
         if stream_data['media_type'] == 'movie' or stream_data['media_type'] == 'episode':
             if plexpy.CONFIG.MOVIE_NOTIFY_ENABLE or plexpy.CONFIG.TV_NOTIFY_ENABLE:
 
+                progress_percent = helpers.get_percent(stream_data['view_offset'], stream_data['duration'])
+
                 for agent in notifiers.available_notification_agents():
                     if agent['on_play'] and notify_action == 'play':
                         # Build and send notification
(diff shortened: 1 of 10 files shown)


plexpy/notification_handler.py
```diff
@@ -43,7 +45,8 @@ def notify(stream_data=None, notify_action=None):
                         # Set the notification state in the db
                         set_notify_state(session=stream_data, state='play', agent_info=agent)
 
-                    elif agent['on_stop'] and notify_action == 'stop':
+                    elif agent['on_stop'] and notify_action == 'stop' \
+                        and (plexpy.CONFIG.NOTIFY_CONSECUTIVE or progress_percent < plexpy.CONFIG.NOTIFY_WATCHED_PERCENT):
```
drzoidberg33, line 49: Shouldn't this be `plexpy.CONFIG.NOTIFY_CONSECUTIVE and progress_percent < plexpy.CONFIG.NOTIFY_WATCHED_PERCENT`?

plexpy/notification_handler.py
```diff
@@ -52,7 +55,8 @@ def notify(stream_data=None, notify_action=None):
 
                         set_notify_state(session=stream_data, state='stop', agent_info=agent)
 
-                    elif agent['on_pause'] and notify_action == 'pause':
+                    elif agent['on_pause'] and notify_action == 'pause' \
+                        and (plexpy.CONFIG.NOTIFY_CONSECUTIVE or progress_percent < 99):
```
drzoidberg33, line 59: Same here.

plexpy/notification_handler.py
```diff
@@ -61,7 +65,8 @@ def notify(stream_data=None, notify_action=None):
 
                         set_notify_state(session=stream_data, state='pause', agent_info=agent)
 
-                    elif agent['on_resume'] and notify_action == 'resume':
+                    elif agent['on_resume'] and notify_action == 'resume' \
+                        and (plexpy.CONFIG.NOTIFY_CONSECUTIVE or progress_percent < 99):
```
drzoidberg33, line 69: And here.

plexpy/notification_handler.py
JonnyWong16, line 49: If notify consecutive checked, it should send regardless of percent.
- if `and` it wouldn't send when progress_percent > watched_percent

If notify consecutive unchecked, it should only send if progress_percent < watched_percent.
- if `and` it would never send

drzoidberg33, line 49: Ah, yes I had the logic backwards.

Issue #219, opened 2015-09-29
[Feature] easier way to filter user history between movies/episodes


tylordurton:
I would like to be able to easily filter the user history between episodes and movies, sounds trivial but my user base watch a LOT of episodes (kids mainly) but would be interested to see what movies are being viewed by simple filtering.

drzoidberg33:
I like what Jonny has done with the breadcrumbs but I know this isn't accessible from the history table directly.

JonnyWong16:
Also accessible from the "Library statistics" on the homepage.

JonnyWong16:
Filtering in history table added in v1.2.4.

PR #220, opened 2015-09-30
Player in tables and top level info pages

JonnyWong16:
- Add "Player" to tables
  - Also rename user platforms to user players
  - Fix other platform vs. player ambiguities
- Add top level info pages with history table
  - Reference #219 - not a filter, but achieves the same result.


drzoidberg33:
Be gone player/platform confusion!

PR #228, opened 2015-10-04
Added ability to set custom notification sound for Pushover

BrianInAz:
User provides desired sounds from official Pushover list. If empty, uses default Pushover sound.

https://github.com/drzoidberg33/plexpy/issues/210


drzoidberg33:
Thanks! I don't use Pushover but the code looks good.

mrchops1024:
I just tested the changes and they seem to be working well. Thanks @BrianInAz!

BrianInAz:
You're welcome!

Ideally I'd like to request the list of sounds from the pushover API and build a drop-down list for the user to select from. The priorities should also use a drop-down. I've not done that in Python yet and couldn't find a good example to work off of. Figured it was best to get this out there so folks can use the sounds.

I'll see about coming back again later and implementing the drop-downs if I find a good example and spare time.

PR #237, opened 2015-10-11
Added in a notifier for IFTTT

rossdargan:
I have _NEVER_ coded in python before, so please do check it over. Also there is an issue with the ssl cert validation. I have had to disable cert verification which is bad news, but I couldn't get it working otherwise. I hate doing a P/R like this, but if you know how to fix it that would be helpful!


drzoidberg33:
Can you submit a new request based off the `dev` branch. Your code is based off `master`.

Hellowlol:
Just a tip since you mention that you have never written anything in python. If your doing something simple http requests you can use requests (already incluced in plexpy)

quick example

```
def notify(self, message, subject):
    header = {}
    data = {}
    url = ''
    r = requests.post(url, header=header, data=data, verify=False)
    if r.status_code == 200:
        result = r.content
        return True
```

rossdargan:
Thanks for the tip, I just copied the code used by the plex notifier (I think!). I'll swap over. 

Is there anything that can be done about ignoring cert verification?

drzoidberg33:
@rossdargan Unfortunately the notifiers code is a bit of a mess at the moment with each notifier sort of doing their own thing.

You can use the `request` method instead, as it uses the global ssl cert override we have in the config.ini.

Hellowlol:
I dont know as i have no idea what error you got. You can try to set verify=True and see what happens. Or use request as @drzoidberg33 mentioned

rossdargan:
I decided not to use the request method. I tried to get it to work, but tbh I found the easier method a bit easier. 

P.S. I don't have the cert issue because I don't validate certs, but thats not a good thing, so my question was why isn't cert validation working.

Hellowlol:
I don't know, what error do you get when you try?

drzoidberg33:
It should work on most systems and you shouldn't need to bypass the cert verification unless you are getting errors.

rossdargan:
This describes the error well https://stackoverflow.com/questions/27835619/ssl-certificate-verify-failed-error

I just couldn't figure out how to fix it, but this is probably the issue I'm having.

drzoidberg33:
What system/OS are you running PlexPy on?

rossdargan:
Windows Server 2008 R2.

drzoidberg33:
Cert checks failing on other services too (like plex.tv)?

rossdargan:
nope, everything else is working fine :/

drzoidberg33:
I just did some tests with this. SSL cert verification works fine on my end.

Also it seems this would be a lot more useful if we sent exactly which event was being triggered (e.g. `plexpy_start`, `plexpy_stop`, `plexpy_pause`, etc). This would require some additional changes to `notification_handler`. I can have a look at doing this if I get some time this evening.

rossdargan:
I'm not sure if it's possible but it would have helped if I could have used some of the notification string substitutes in the settings. I.e. for the event on the IFTTT notifications could be {player} that would allow me to say turn a light in the lounge on when I pause a program only on my lounge player.

The code to do the string substitution must already exist, but because we don't know what generated the notification I don't think it would be very easy at the minute to use it.

Coming from C# this is all a bit of a challenge for me anyway!

BTW we should close this PR and switch to the one I did from the dev branch

drzoidberg33:
Closing this PR.

PR #245, opened 2015-10-13
Raise exception so the uri with the xml

Hellowlol:
that fails are logged.
- this is untested btw


drzoidberg33:
I don't think this is the way to go, it's inconsistent. I haven't looked but there must be a cleaner way to achieve this.

Hellowlol:
That is the entire point. We cant use the helper as it kills all exceptions and I didn't want to re raise from it as other code depends on it.

 I don't know any other way to solve this, but the URL should be logged someway, guessing what xml might have failed is no fun :p

PR #249, opened 2015-10-15
Capitalise "Python"

dmptrluke:
This is a super minor and pointless thing, but it seem that in official documentation Python is capitalised, so thought I'd suggest using it here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Join the chat at https://gitter.im/drzoidberg33/plexpy](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/drzoidberg33/plexpy?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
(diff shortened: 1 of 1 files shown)


drzoidberg33:
All PRs should be made to the dev branch.

Hellowlol:
Alot of prs are getting closed since they are not to dev. I think it would nice if it also stated that all prs should be sendt to dev. (I know its mentioned in the branches paragraph, but i didnt read that, and suspect other are doing the same)

drzoidberg33:
The info is in the correct place, I believe it also comes up when making the PR to make sure you've read the contributions document.

screenshot 2015-10-15 10 53 56

Hellowlol:
Yes, but that doesnt help as long as the info is in the branches header. That info should be moved to the https://github.com/drzoidberg33/plexpy/blob/master/CONTRIBUTING.md#pull-requests.

Under branches, it should state that master is "stable" and that the brave and bold can run dev to get the latest and the greatest.

Anyway, its up too you. Just thought it was worth mentioning since this happens on a weekly basis (from the time i started to watch your repo)

PR #256, opened 2015-10-19
Miscellaneous fixes

JonnyWong16:
Resubmit PR to dev branch.

Reference #219.

### DIFF
--- a/plexpy/datatables.py
+++ b/plexpy/datatables.py
@@ -70,6 +70,39 @@ class DataTables(object):
         else:
             grouping = False
 
+        # Build join parameters
+        if join_types:
+            counter = 0
+            for join_type in join_types:
+                if join_type.upper() == 'LEFT OUTER JOIN':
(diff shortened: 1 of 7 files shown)


plexpy/datatables.py
```diff
@@ -70,6 +70,39 @@ def ssp_query(self,
...
+                c_where += w[0] + ' = ? AND '
+
+                # The order of our args changes if we are grouping
+                #if grouping:
+                #    args.insert(0, w[1])
+                #else:
+                #    args.append(w[1])
+
+                # My testing shows that order of args doesn't change
+                args.append(w[1])
```
JonnyWong16, line 101: Need to confirm this.

I think it should be okay since custom where args are created before the search filter args now.

drzoidberg33, line 101: I honestly can't even remember why I needed to do this. I believe it had to do with custom searches and grouping.

drzoidberg33:
Looks good.

PR #257, opened 2015-10-19
Add library_id

JonnyWong16:
Commits contain #256 since I branched off that.

### DIFF
--- a/plexpy/__init__.py
+++ b/plexpy/__init__.py
@@ -609,15 +609,6 @@ def dbcheck():
         c_db.execute(
             'ALTER TABLE session_history_metadata ADD COLUMN library_title TEXT'
         )
-        logger.debug(u"Updating library_id's in database. Please wait...")
-        from plexpy import datafactory
-        data_factory = datafactory.DataFactory()
-        result = data_factory.update_library_ids()
-
-        if result:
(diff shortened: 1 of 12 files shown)


plexpy/__init__.py
```diff
@@ -547,13 +598,26 @@ def dbcheck():
...
+        c_db.execute(
+            'ALTER TABLE session_history_metadata ADD COLUMN library_id INTEGER'
+        )
+        c_db.execute(
+            'ALTER TABLE session_history_metadata ADD COLUMN library_title TEXT'
+        )
+        logger.debug(u"Updating library_id's in database. Please wait...")
+        from plexpy import datafactory
+        data_factory = datafactory.DataFactory()
+        result = data_factory.update_library_ids()
```
drzoidberg33: Maybe make sure the server is online and that we are connected to the correct server before running this.

This should preferably be invoked manually by the user.

JonnyWong16:
This should be better.

drzoidberg33:
@JonnyWong16 I added some things to this PR, can you check over at #263. I'm going to close this PR and we can work off that branch.

Fixed the PlexWatch importer (you seem to be hell bent on breaking it :laughing:) and a few other small bugs as well as giving the library sections their own table.

PR #259, opened 2015-10-21
Recently added

JonnyWong16:
Finally attempting this.

Proof of concept and it's still a mess (especially `notification_handler`). Only works with web sockets for now. Not sure how this will work with the activity pinger since the `/library/recentlyAdded` endpoint doesn't provide episodes/tracks.

I need some testing and feedback.


drzoidberg33:
Woop. Thanks for taking this on.

If you want you can see how PlexWatch does this but I honestly don't mind if it's restricted to websocket-only feature. Ideally websockets will be the default method going forward.

Only issue that looks like it still needs to be addressed is buffering the recently added events otherwise if somebody adds a bunch of items at once they're going to be spammed up the wazoo. I'll see if I can assist with this when I get a moment.

Arcanemagus:
Just an FYI PlexWatch sends a notification for each newly added item, which is how I at least prefer it to happen.

drzoidberg33:
Thanks @Arcanemagus - I think this is something you should be able to configure at least. I certainly wouldn't want to be notified of every single item if I add a new batch of media.

plexpy/notification_handler.py
```diff
@@ -293,24 +310,26 @@ def build_notify_text(session, state):
         full_title = session['title']
 
     # Generate a combined transcode decision value
-    if session['video_decision']:
+    transcode_decision = ''
+    if session.get('video_decision', None):
```
drzoidberg33: I don't think `.get` works with sqlite objects.

```
Traceback (most recent call last):
  File "/Users/tim/PyCharmProjects/plexpy-jw/plexpy/logger.py", line 210, in new_run
    old_run(*args, **kwargs)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/Users/tim/PyCharmProjects/plexpy-jw/plexpy/notification_handler.py", line 128, in notify
    notify_strings = build_notify_text(session=stream_data, state=notify_action)
  File "/Users/tim/PyCharmProjects/plexpy-jw/plexpy/notification_handler.py", line 314, in build_notify_text
    if session.get('video_decision', None):
AttributeError: 'sqlite3.Row' object has no attribute 'get'
```

Any reason for this change?

Hellowlol:
Should be easy to just find get the recently added. You can use this as a example for nonsocket method. https://github.com/Hellowlol/HTPC-Manager/blob/master2/modules/plex.py#L129

Arcanemagus:
@drzoidberg33 Actually, that's a really good point. Grouping things like "Season X added" would be cool, but I'm not sure what you would use as thresholds for that.

plexpy/notification_handler.py
JonnyWong16: I'll change it to `if 'video_decision' in session:`.

I need some way to differentiate between session notifications and timeline notifications. I'll clean up `notification_handler`.

JonnyWong16:
@Hellowlol Thanks, that'll help.

JonnyWong16:
What's your opinion on logging the `on_created` notification in the database? Unnecessary?

drzoidberg33:
@JonnyWong16 Yes, we should log.

And sorry I screwed up a clean merge with the new IFTTT notifier.

JonnyWong16:
Alright, ready for testing again.

JonnyWong16:
Alright, ready for testing again.

drzoidberg33:
Thanks, will give it another spin today.

data/interfaces/default/settings.html
```diff
@@ -502,6 +502,12 @@
                             </label>
                             <p class="help-block">Disable to prevent consecutive notifications (i.e. both watched &amp; stopped notifications).</p>
                         </div>
+                        <div class="checkbox">
+                            <label>
+                                <input type="checkbox" name="notify_recently_added_grandparent" id="notify_recently_added_grandparent" value="1" ${config['notify_recently_added_grandparent']}> Group notifications for recently added TV Shows or Music
+                            </label>
+                            <p class="help-block">Endable to only get one notification for recently added Episodes or Tracks. Movies are unaffected.</p>
```
drzoidberg33: Small typo :smile:

data/interfaces/default/info.html
```diff
@@ -11,7 +11,7 @@
 
 == Global keys ==
 rating_key              Returns the unique identifier for the media item.
-type                    Returns the type of media. Either 'movie', 'show', 'season', 'episode', 'artist', 'album', or 'track'.
+meida_type                    Returns the type of media. Either 'movie', 'show', 'season', 'episode', 'artist', 'album', or 'track'.
```
Hellowlol: typo

drzoidberg33:
@JonnyWong16 

I've tested what I can and seems to be working.

Just fix up those typo's and then you can merge this PR to dev (you should be able to now).

I run the dev branch on my main Plex machine so will test for a few days before we release to master.

PR #265, opened 2015-10-27
Change IP lookup provider

### LINKED ISSUE #264: [Task] Change IP lookup service to provider that supplies SSL enabled endpoint.
Running PlexPy behind an apache2 reverseproxy, access to apache2 interface is via HTTPS.

IP-API lookups are done at the client side rather than by the PlexPy app serverside, and HTTP requests from HTTPS pages are blocked by browser security:

[blocked] The page at https://myserver.com/ppy/history was not allowed to display insecure content from http://ip-api.com/json/1.2.3.4.

drzoidberg33:
Resolves #264.

Adds SSL enabled IP lookup service.


JonnyWong16:
I would just switch the order so it's Country > Region > City.

Issue #51, opened 2015-08-11
New notification agent: Telegram Bot - feature request


elRadix:
There seems to be a project where plex can interact with telegram. I would be nice if it can be integrated

The user needs to create his own bot and generate his personal telegram token

https://github.com/brownsmart/telegram-plexbot
https://forums.plex.tv/discussion/176090/new-app-telegrambot-plex-api#latest

Thank you for looking into it !

elRadix:
@drzoidberg33 @JonnyWong16 @Hellowlol 

There is a simple script created by a user to send telegram notifications
https://gist.github.com/matriphe/9a51169508f266d97313

you only need a userid and a botkey, can you integrate this ?

would be also nice to send a cover of movie/serie optionally ...

onedr0p:
This has been implemented in the pull request below.

elRadix:
would it be possible to add sending the cover photo of the media from plex to your telegram bot ?
you can do this in the Notification string substitutions, so that each user can choose what he wants to add.

JonnyWong16:
Telegram added in v1.2.4.

Request for posters in notifications in #273.

PR #267, opened 2015-10-28
Feature/telegram notifications

onedr0p:
This is a feature request filled for #51 @drzoidberg33 If you can, please merge in with current dev branch because that has the on_created notifications. Thanks!

FYI, I choose to use the [leandrotoledo/python-telegram-bot](https://github.com/leandrotoledo/python-telegram-bot/blob/master/examples/roboed.py) library in case of future updates to the Telegram API. Let me know if this is a problem.


drzoidberg33:
Thanks for the PR. I'll try and review this a bit later today.

onedr0p:
Thanks for your continued support on this project, I'll see what I can do to help.

drzoidberg33:
Sorry, I made comments on your fork by accident, but check those out. Otherwise it looks good.

I'm also wondering if we need such a massive Telegram library to just send messages, is there anything more lightweight that we can use?

drzoidberg33:
I'm wondering now if there is any reason to even include that library. We're using one endpoint, which is a simple HTTP request.

We'll also be able to catch and log our errors more cleanly.

Anything I'm missing?

onedr0p:
Telegram seems to be pretty active in their development, I included the library in case they change their api. That and it was easier to put together. I can look at adding support without the lib if your interested, or look for a more lightweight one.

onedr0p:
I took your advice and made it work without the lib, it just uses a simple HTTP request.

I tested it for a bit and didn't see any problems.

drzoidberg33:
Sorry, been super busy and haven't had time to review the new changes yet. Will take a look tonight hopefully.

plexpy/notifiers.py
```diff
@@ -1417,3 +1437,68 @@ def return_config_options(self):
...
+        return cherrypy.config['config'].get('Telegram', options)
+
+    def notify(self, message, event):
+        if not message or not event:
+            return
+
+        http_handler = HTTPSConnection("api.telegram.org")
+
+        data = {'chat_id': self.chat_id,
+                'text': message.encode("utf-8")}
```
drzoidberg33: We're dropping the message subject here, can't we do something like with the Twitter notifications and format it like:

Subject: Message

onedr0p: I think I understand.

``` python
elif config_id == 13:
          telegramClient = TELEGRAM()
          telegramClient.notify(message=body, event=subject)
```

I pass the subject as event, so is this what you mean?

``` python
data = {'chat_id': self.chat_id,
    'text': event + ': ' + message.encode("utf-8")}
```

onedr0p: Updated the PR to include this.

plexpy/notifiers.py
```diff
@@ -1417,3 +1437,68 @@ def return_config_options(self):
...
+        return cherrypy.config['config'].get('Telegram', options)
+
+    def notify(self, message, event):
+        if not message or not event:
+            return
+
+        http_handler = HTTPSConnection("api.telegram.org")
+
+        data = {'chat_id': self.chat_id,
+                'text': event + ': ' + message.encode("utf-8")}
```
drzoidberg33: `event` should be encoded too.

onedr0p: updated in PR

drzoidberg33:
Thanks! Merging.

onedr0p:
sweet, looking forward to the new release!

PR #278, opened 2015-11-11
Add IP address to current activity

JonnyWong16:


drzoidberg33:
We can make `ip_address` a notification substitution too now.

JonnyWong16:
It will still show up as "N/A" for `state=playing` for anyone not on 0.9.14 though.

drzoidberg33:
That's fine, they will eventually be on 0.9.14 anyway. Or what you can do is make it available but don't document it in the info modal until 0.9.14 is public.

Arcanemagus:
Ooo, am I reading this right and they _finally_ put the IP address in the session info in 0.9.14?

drzoidberg33:
Looks good :shipit:

Issue #281, opened 2015-11-13
[Feature] Option to add server name to browser window title


pax0707:
For us who have multiple servers it would be nice to be able to add server name to tab title to easily identify the tab for each server.

Hellowlol:
https://github.com/drzoidberg33/plexpy/pull/282

JonnyWong16:
Added in v1.2.4.

Issue #279, opened 2015-11-12
Hybrid IP notation 


pax0707:
It's providing " Invalid IP Address" for location feature.

 12.11.2015.    pax0707 ::ffff:192.168.0.10 iOS 
 Pax
 Monty Python and the Holy Grail (1975)
20:17   0 mins  20:20   3 mins  

 12.11.2015.
ursuea  ::ffff:212.15.178.89    iOS 
 Tigana
 The Man from U.N.C.L.E. (2015)
20:17   0 mins  20:21   4 mins

JonnyWong16:
IPv6 added in v1.2.4.

Issue #273, opened 2015-11-07
Feature Request: Add genre and poster url to notifications


zobe123:
Thank you guys for the recently added feature. Great work. Can you also add the keywords "genre" and a poster-link? Thx

JonnyWong16:
Genre is okay, but a poster url would be messy.

zobe123:
how would the recommended practice be, to insert posters into email?

I'm the only one where the TV Tag is not working properly? (Movie Tag all fine)

...is it possible to highlight each Tag? (Feature Request?)
...is it possible to get more than one line for the code? (Feature Request?)

![11 11](https://cloud.githubusercontent.com/assets/13840542/11089484/5c572d4a-886a-11e5-805b-49977d767473.png)

drzoidberg33:
@zobe123 Please open up your own issue for feature requests (note that multi-line has been implemented already.)

PR #282, opened 2015-11-15
IPv6 support, fix recently added, get server name, notify server down

JonnyWong16:
Reference #279, #273, #281.


plexpy/webserve.py
```diff
@@ -46,9 +46,13 @@ def serve_template(templatename, **kwargs):
 
     _hplookup = TemplateLookup(directories=[template_dir])
 
+    # Get the server name
+    pms_connect = pmsconnect.PmsConnect()
+    server_name = pms_connect.get_server_friendly_name()
```
drzoidberg33: I don't particularly like this, it means we have to wait for a server response before we can load our page. If the server is down or slow to respond every time you hit the page it's going to be slow.

I'd rather store this in the config file and refresh it on start up or with the scheduler in the background.

drzoidberg33: I don't particularly like this, it means we have to wait for a server response before we can load our home page. If the server is down or slow to respond every time you hit the page it's going to be slow.

I'd rather store this in the config file and refresh it on start up or with the scheduler in the background.

drzoidberg33:
Lots of great stuff in here, give me a few hours to go through all of it.

plexpy/__init__.py
```diff
@@ -280,14 +281,19 @@ def initialize_scheduler():
...
 
         if CONFIG.PMS_IP and CONFIG.PMS_TOKEN:
-            schedule_job(plextv.get_real_pms_url, 'Refresh Plex Server URLs', hours=12, minutes=0, seconds=0)
+            schedule_job(plextv.get_real_pms_url, 'Refresh Plex Server URLs',
+                         hours=12, minutes=0, seconds=0)
+            schedule_job(pmsconnect.get_server_friendly_name, 'Refresh Plex Server Name',
+                         hours=12, minutes=0, seconds=0)
+            schedule_job(activity_pinger.check_recently_added, 'Check for recently added items',
+                         hours=0, minutes=0, seconds=seconds)
+            schedule_job(activity_pinger.check_server_response, 'Check for server response',
```
drzoidberg33, line 290: Thinking out loud here but why do we need to add another task to do this? We're already pinging the server to check for activity, so why not just implement it there?

For example, in activity_pinger:

```
if session_list:
    do stuff
else:
    server_connected = check_server_connectivity
    if not server_connected:
        trigger server_down
```

Those with websocket connections will fail over to the pinger, so this would work for both.

JonnyWong16, line 290: Very good point....

JonnyWong16:
I think I have to move the check back to it's own task, so I can check if we get an internal response (`ip:port/endpoint` is down) or an external response (`plex.tv/pms/servers.xml` returns empty xml if the port is down). It would just be much simpler this way, instead of adding a check for external response in the same function.

drzoidberg33:
I just don't see any place to disable the online check. Looks like it's just using the default monitoring interval but if you don't have remote access enabled on your system then it's going to spam both the server and the PlexPy log unnecessarily.

drzoidberg33:
Cool. Let's merge this into dev (if you're happy).

drzoidberg33:
Thanks! I will test out on my side for a few days but want to push a release as soon as possible.

drzoidberg33:
Getting false positives on remote reachability.

```
2015-11-21 17:07:55 WARNING PlexPy Monitor :: Unable to get an external response from the server, ping attempt 25.
2015-11-21 17:05:55 WARNING PlexPy Monitor :: Unable to get an external response from the server, ping attempt 24.
2015-11-21 17:03:55 WARNING PlexPy Monitor :: Unable to get an external response from the server, ping attempt 23.
2015-11-21 17:01:55 WARNING PlexPy Monitor :: Unable to get an external response from the server, ping attempt 22.
2015-11-21 16:59:55 WARNING PlexPy Monitor :: Unable to get an external response from the server, ping attempt 21.
2015-11-21 16:57:55 WARNING PlexPy Monitor :: Unable to get an external response from the server, ping attempt 20.
2015-11-21 16:55:55 WARNING PlexPy Monitor :: Unable to get an external response from the server, ping attempt 19.
```

Not sure why.

I see you're not using the method Plex Web uses to check remote access reachability, maybe that's something to look into. There are two urls used there:

https://the_server_url:32400/myplex/refreshReachability
https://the_server_url:32400/myplex/account

drzoidberg33:
Also disabling the check requires you to restart to get it to take effect. You need to reset the scheduler when changing this.

mbnn:
Disabling Music notifications doesnt work for me, i still get them.
Basically only need TV and movies recently added.

JonnyWong16:
@drzoidberg33 
I get `404 Not Found` for `/myplex/refreshReachability`

@mbnn 
Those checkboxes don't affect recently added notifications.

drzoidberg33:
@JonnyWong16 I see it's a PUT request.

Just check the network requests from Plex Web in the Server settings page.

JonnyWong16:
@drzoidberg33 
~~I see it now, but not sure how to use it.~~

Never mind, it refreshes the port mapping. See #289.

PR #289, opened 2015-11-21
Use Plex API to check remote access down

JonnyWong16:
This should be better now.


JonnyWong16:
I'm merging into dev so we can test.

drzoidberg33:
Thanks, I'm updating my install.

drzoidberg33:
Bit of a derp on the first run wizard:

screenshot 2015-11-23 11 16 25

data/interfaces/default/settings.html
```diff
@@ -405,9 +405,9 @@
                         </div>
                         <div class="checkbox">
                             <label>
-                                <input type="checkbox" id="monitor_remote_access" name="monitor_remote_access" value="1" ${config['monitor_remote_access']}> Monitor Plex Remote Access
+                                <input type="checkbox" class="monitor-settings" id="monitor_remote_access" name="monitor_remote_access" value="1" ${config['monitor_remote_access']}> Monitor Plex Remote Access
```
drzoidberg33, line 408: This doesn't require a restart.

Just need to restart the scheduler:

`plexpy.initialize_scheduler()`

data/interfaces/default/settings.html
```diff
@@ -479,7 +479,12 @@
                         </div>
                         <div class="checkbox">
                             <label>
-                                <input type="checkbox" name="movie_notify_enable" id="movie_notify_enable" value="1" ${config['movie_notify_enable']}> Enable Movie and TV Notifications
+                                <input type="checkbox" name="movie_notify_enable" id="movie_notify_enable" value="1" ${config['movie_notify_enable']}> Enable Movie Notifications
+                            </label>
+                        </div>
+                        <div class="checkbox">
+                            <label>
+                                <input type="checkbox" name="tv_notify_enable" id="tv_notify_enable" value="1" ${config['tv_notify_enable']}> Enable TV Show Notifications
```
drzoidberg33, line 487: You need to enable this after an update if the user had the old global video notifications option enabled. After updating on my system it disabled TV notifications, people going to wonder why their notifications are broken.

data/interfaces/default/settings.html
JonnyWong16, line 408: Actually it doesn't require either?

The scheduled task already runs, and the setting just toggles that if statement.

data/interfaces/default/settings.html
JonnyWong16, line 487: ~~Not sure where or how I would do this.~~

The original Headphones code uses `CONFIG_VERSION` to keep track of upgrades. I'll try that.

PR #306, opened 2015-11-28
Delete users

JonnyWong16:
Purges history and sets `deleted_user` to `1` in the `users` table so it won't reappear after a user refresh.

Can re-enable the user through `/undelete_user?&user_id=<user_id>` or `/undelete_user?username=<username>`.

The user still shows up in the current activity. Not sure what the best way is to handle this.


drzoidberg33:
Hey, sorry been busy with so much other stuff lately. Will take a look today.

drzoidberg33:
Looks good to me.

Issue #319, opened 2015-12-02
Recently added calls being made even when disabled


drzoidberg33:
The recently added job is doing a lot of calls to PMS, if we don't have recently added enabled it shouldn't be making these calls.

JonnyWong16:
Check if any notifier has `on_created` enabled? or implement another global toggle for recently added?

drzoidberg33:
I'd prefer (and would be easier) to include a global toggle.

JonnyWong16:
Should be fixed in v1.3.0.

Issue #318, opened 2015-12-02
Making calls for server availability even when disabled


drzoidberg33:
I have remote monitoring disabled and no notification triggers enabled yet on every monitor call PlexPy is still making requests for availability.

```
02-Dec-2015 22:03:07 - DEBUG :: Thread-14 : PlexPy HTTP Handler :: [PUT] http://192.168.1.3:32400/myplex/refreshReachability
02-Dec-2015 22:03:15 - DEBUG :: Thread-14 : PlexPy HTTP Handler :: [GET] http://192.168.1.3:32400/myplex/account
```

JonnyWong16:
The setting "Monitor Plex Remote Access" toggles the check for the port mapping. The check for a server response runs regardless of that setting.

I guess the setting should be changed to include both checks.

drzoidberg33:
The regular server up/down check shouldn't need to poll those URLs. The refreshReachability endpoint causes PMS to talk to plex.tv every time, if we don't need this it should be disabled.

drzoidberg33:
Cool, I believe this is resolved.

Thanks Jonny.

PR #324, opened 2015-12-04
Only schedule job for recently added or monitor remote access if sett…

JonnyWong16:
…ing enabled
#318, #319

bb5aa2b significantly reduces the number of `/metadata` calls.

### DIFF
--- a/data/interfaces/default/settings.html
+++ b/data/interfaces/default/settings.html
@@ -313,7 +313,8 @@ available_notification_agents = sorted(notifiers.available_notification_agents()
                                     <input type="text" class="form-control" id="pms_logs_folder" name="pms_logs_folder" value="${config['pms_logs_folder']}" size="30" data-parsley-trigger="change">
                                 </div>
                             </div>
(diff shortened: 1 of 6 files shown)


drzoidberg33:
Excellent. Going to test this out.

drzoidberg33:
Cool this helps a lot. Integrating.

drzoidberg33:
Seems I was a bit hasty to merge this.

Recently added no longer works at all. I took a look at what's happening and there are two main issues I picked up:
1. There is no time for the `RECENTLY_ADDED_DELAY` to do it's thing. The list of recently added items is cleared out if the previous list was the same so all that code is skipped.
2. When the last item we added is in the same show/season as the new one we're adding it will ignore it because recently added works off the season rating key. Rather use `lastUpdated` to check if there were changes.

mbnn:
So does this now mean we can just Enable Recently Added Notifications but leave the others disabled to simply only push out recently added notifications? Or do the other options also need to stay enabled (Enable Movie Notifications/Enable TV Show notifications etc?).

JonnyWong16:
Currently, you need both checked. I plan on changing this to notify by library section instead of by media type in the future.

PR #325, opened 2015-12-04
Update base.html

realvz:
Current-Activity Header is hidden in Mid and small devices because the logo is too big and it causes Navigation bar to appear on top of the activity header or hides show->episodes navigation.

This way, it will still show full logo on larger display and will show smaller logo on smaller display so information is not hidden.


drzoidberg33:
Please can you resubmit this to the `dev` branch. Thanks!

PR #326, opened 2015-12-04
Update base.html

realvz:
Current-Activity Header is hidden in Mid and small devices because the logo is too big and it causes Navigation bar to appear on top of the activity header or hides show->episodes navigation.

This way, it will still show full logo on larger display and will show smaller logo on smaller display so information is not hidden.

### DIFF
--- a/data/interfaces/default/base.html
+++ b/data/interfaces/default/base.html
@@ -51,8 +51,11 @@ from plexpy import version
                     <span class="icon-bar"></span>
                     <span class="icon-bar"></span>
                 </button>
-                <a class="navbar-brand" href="home">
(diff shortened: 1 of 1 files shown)


JonnyWong16:
I think I'd rather just switch to the collapsed menu for mid-sized screens.

I might add more tabs in the future and the navbar would run out of space even with the smaller logo.

realvz:
I would hate to have collapsed menu on iPad. I frequently navigate between graphs history and home. Perhaps there's happy medium where main buttons exist and a collapsed menu for items that are not used frequently. To take it even further, user can chose the 5 (or 4 or 6) menu items to be shown as buttons, kind of like iPod app used to do in ios8.

JonnyWong16:
Closing. This should be better with #367.

PR #328, opened 2015-12-05
Fix recently added

JonnyWong16:
Resolve issues from #324

I'm sure there was a reason, but I don't remember why I didn't do it this way originally.


JonnyWong16:
I'm merging so I can push a release with sanitized player names.

PR #353, opened 2015-12-08
Get watched status

Hellowlol:
Its just to show how i grabbed the info.


Hellowlol:
I just noticed that i forgot to add requests_futures.

Hellowlol:
@JonnyWong16 Would something like this work? Or had you intented something completely different.

JonnyWong16:
I haven't had time to look.

plexpy/pmsconnect.py
```diff
@@ -692,11 +697,225 @@ def get_metadata_children_details(self, rating_key=''):
...
+            port = plexpy.CONFIG.PMS_PORT
+            self.protocol = 'http'
+
+        b_url = '%s://%s:%s' % (self.protocol, hostname, port) 
+
+        h = {'Accept': 'application/json',
+             'X-Plex-Token': plexpy.CONFIG.PMS_TOKEN
+            }
+
+        hide_watched = 'unwatched' if hide_watched else 'all'
```
JonnyWong16, line 746: I believe `/library/sections/<key>/unwatched` only lists unwatched for the logged in user (i.e. yourself)?

Something could be watched by someone else, and we don't want it to show up as unwatched.

plexpy/pmsconnect.py
```diff
@@ -692,11 +697,225 @@ def get_metadata_children_details(self, rating_key=''):
...
+                            'lastViewedAt': item.get('lastViewedAt', 0),
+                            'viewOffset': item.get('viewOffset', 0),
+                            'spaceUsed': 0, # custom 
+                            'isSeen': 0, # custom
+                            '_elementType': item['_elementType']
+                        }
+
+                        # Only movies has the files listed here
+                        if item['_elementType'] == 'Video':
+                            d['viewCount'] = item.get('viewCount', 0)
```
JonnyWong16, line 797: Same as above, view count is only for yourself, not over all users.

Total views should be pulled from the PlexPy db.

plexpy/pmsconnect.py
```diff
@@ -692,11 +697,225 @@ def get_metadata_children_details(self, rating_key=''):
...
+
+                        #elif item['_elementType'] == 'Track':
+                        #    # I dont think it returns a "Track" as all
+                        #    pass
+
+                        elif item['_elementType'] == 'Directory':
+
+                            if item['type'] == 'show' or item['type'] == 'artist':
+                                d['viewedLeafCount'] = item.get('viewedLeafCount', 0)
+                                d['leafCount'] = item.get('leafCount', 0)
```
JonnyWong16, line 824: Same again.

plexpy/pmsconnect.py
```diff
@@ -692,11 +697,225 @@ def get_metadata_children_details(self, rating_key=''):
...
+            try:
+                v = int(v)
+            except:
+                try:
+                    v = float(v)
+                except:
+                    pass
+            return v
+
+        result = fetch('/library/sections/all').get('_children')
```
JonnyWong16, line 763: Any reason we can't use the existing `request_handler`?

plexpy/pmsconnect.py
```diff
@@ -692,11 +697,225 @@ def get_metadata_children_details(self, rating_key=''):
...
+                                    f_result.append(d)
+                                                    
+                            #elif item['type'] == 'artist':
+                            #    pass # Handled above left for future stuff
+                        
+
+        t_result = f_result
+
+        if get_file_size:
+            future_sess = FuturesSession(max_workers=8)
```
JonnyWong16, line 849: What does this `future` section do?

JonnyWong16:
I'm thinking about how we can implement this with Datatables.

Also, I think we can speed this up significantly using paging `X-Plex-Container-Size` and `X-Plex-Container-Start`. Only pull enough metadata to populate one page of a Datatable at a time.

plexpy/pmsconnect.py
Hellowlol, line 849: It fires the http requests async greatly increasing the speed.

plexpy/pmsconnect.py
Hellowlol, line 763: I dont want any console logging slowing down the process and it was a cleaner way to do it imo

Hellowlol:
Isnt the idea to grab every item? :P

I added hide_wathed since i assumed it would be much faster if pms handled this, but it uses twice as long time.

If you want to get the all unwatched content the quickest way to to do: http://localhost:8181/get_watched?get_file_size=0&isSeen=0
using 1,50 ish sec.

If we want it faster we can catch the inital response for xxx minutes then do the filtering based on the kwargs

plexpy/pmsconnect.py
Hellowlol, line 746: Ah, that might be i have never really used it :D

plexpy/pmsconnect.py
Hellowlol, line 797: How it it possible to find stuff to find content that hasnt been watced by anybody? :(

drzoidberg33:
I haven't been following this much. What exactly are we trying to do here?

Hellowlol:
@drzoidberg33 The pr was about getting info about unwatched videos/episodes. But i failed to realize that we have to get the watched info from the db, so the entire pr was flawed since you can only see the watched info from the owner om the pms.

Very much of the pr is reusable. We just have to compare the a list dicts with ratingkeys from the db again a list of dict ratingkey pulled from pms.

JonnyWong16:
Yes grab every item, but if we are only displaying 25/50/100 items per page on the Datatable we don't need to retrieve potentially 1000s of items from the API.

I can't think of any filtering we need, just sorting by the different columns we use on the table.

PR #363, opened 2015-12-09
add .pem cert to ignore list

InAnimaTe:
A lot of people generate self-signed certs and use the .pem format...lets make sure those never make it back upstream!


drzoidberg33:
Thanks.

PR #365, opened 2015-12-10
Update current_activity_header.html

zobe123:
dynamic "s" at the word stream/streams
added "no" before Activity when their is no activity

### DIFF
--- a/data/interfaces/default/current_activity_header.html
+++ b/data/interfaces/default/current_activity_header.html
@@ -15,11 +15,15 @@ DOCUMENTATION :: END
 </%doc>
 
 % if data != None:
-% if data == '0':
-<h3>Activity</h3>
+	% if data == '0':
+	<h3>no Activity</h3>
+	% else:
+		% if data == '1':
+		<h3>Activity <small>${data} stream</small></h3>
+		% else:
+		<h3>Activity <small>${data} streams</small></h3>
+		% endif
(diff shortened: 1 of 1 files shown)


zobe123:
Maybe there is no need to add the "no" ?!

JonnyWong16:
Yeah, no need for the "no" since it already says "Nothing is currently being watched" beneath.

data/interfaces/default/current_activity_header.html
```diff
@@ -15,11 +15,15 @@
...
 % if data != None:
-% if data == '0':
-<h3>Activity</h3>
-% else:
-<h3>Activity <small>${data} stream(s)</small></h3>
-% endif
+	% if data == '0':
+	<h3>Activity</h3>
+	% else:
+		% if data == '1':
```
JonnyWong16, line 21: You can simplify by using else if:

```
% elif data == '1':
<h3>Activity <small>${data} stream</small></h3>
% else:
<h3>Activity <small>${data} streams</small></h3>
% endif
```

zobe123:
i didn't know enough about the language you used at this Point. I only copy/pasted the working code... feel free to do the changes you want to have.

JonnyWong16:
For your future reference cb7ba7f, and the language is Mako.

zobe123:
thx

PR #367, opened 2015-12-11
Fix greedy search bar.

drzoidberg33:
@JonnyWong16 Could you take a look to see if these css changes are right, not sure if all those changes were necessary.


JonnyWong16:
Reduce search bar width and hide on small screens?

There's also the width in the javascript at the bottom of `base.html` (lines 143 and 148).

drzoidberg33:
Yep, the current width causes the nav bar to overlap on smaller screens (like an iPad). I'll check that js, thanks.

zobe123:
Yeah, works like a charm now. Thx!
Btw, on ipad it's also impossible to manage the cards options under homepage statistics. It looks like they inactive.

PR #380, opened 2015-12-17
Added Statusbar hover effect with percentage

zobe123:


JonnyWong16:
Cool, thanks.

I just want to add a slight gradient to the bar so we can see the white text.

PR #390, opened 2015-12-21
CentOS 6.X startup script

boopzz:
Init file for CentOS 6.X systems, variables exist for different install directorys. Please note, current version of Python is 2.6 and PlexPy requires 2.7. A variable exists to point to this path.

Since PlexPy is based on Headphones, I just copied the init script for Headphones and changed the paths and added the Python2.7 path


JonnyWong16:
Please resubmit to the Dev branch. Thanks.

Issue #413, opened 2015-12-30
Feature Request: Send one-off custom messages to notification agents


krohrsb:
I love all the notification abilities with PlexPy. However it may be useful to allow sending custom one off messages through the user interface.

I realize there is plex server up/down notifications. But what if I wish to send something specific? Such as   notifying a telegram chat about maintenance or server issues, or general broadcast type messages?

I envisioned it as simply a Text Box in the UI that i can type in whatever and hit "send" to send it to notification agents that enable this 'custom message' type notification.

If it's out of scope or too crazy feel free to say so! Thanks.

Hellowlol:
Dunno about a text field, but i intend to add a notifications to the api :)

JonnyWong16:
I was thinking of adding a "Test notification" button to each of the agents. I could possibly just have a custom textbox for the "Test" notification.

krohrsb:
Yeah that would work for my current needs I believe. I just need to provide custom text on demand occasionally

JonnyWong16:
Added in v1.3.0.

Issue #411, opened 2015-12-29
Pushover Test Event


bonyboy:
Is it possible to add a test feature for this Notification Agent?  This feature is available for other Agents, but not Pushover.

JonnyWong16:
Added in v1.3.0.

Issue #396, opened 2015-12-24
Pushover API Limit hit


erunion:
```
23-Dec-2015 16:16:00 - INFO :: Thread-13 : Pushover request failed: 
23-Dec-2015 16:16:01 - DEBUG :: Thread-14 : Pushover response status: 429
23-Dec-2015 16:16:01 - DEBUG :: Thread-14 : Pushover response headers: [('status', '429'), ('x-request-id', '70dcfa5c0daa2c03fcc46545f9960c86'), ('access-control-allow-headers', 'X-Requested-With, X-Prototype-Version, Origin, Accept, Content-Type, X-CSRF-Token, X-Pushover-App'), ('access-control-max-age', '1728000'), ('transfer-encoding', 'chunked'), ('x-limit-app-reset', '1451628000'), ('x-runtime', '0.004977'), ('connection', 'keep-alive'), ('x-ua-compatible', 'IE=Edge,chrome=1'), ('cache-control', 'no-cache'), ('date', 'Thu, 24 Dec 2015 00:16:01 GMT'), ('access-control-allow-origin', '*'), ('access-control-allow-methods', 'POST, OPTIONS'), ('content-type', 'application/json; charset=utf-8'), ('x-limit-app-remaining', '0'), ('x-limit-app-limit', '7500')]
23-Dec-2015 16:16:01 - DEBUG :: Thread-14 : Pushover response body: '{"token":"invalid","errors":["application has exceeded current limit of 7500 messages"],"status":0,"request":"70dcfa5c0daa2c03fcc46545f9960c86"}'
```

I can easily fix this on my own install by setting up my own API key with Pushover, but this is probably worth reporting anyways.

Maybe we should just remove the hardcoded API key, and force people to put in their own? Bit more barrier to entry with this notifier, but I'm sure the limit will get hit again, resulting in continued failures.

JonnyWong16:
We know. The 7500 notifications get used up halfway through each month.

erunion:
Ah, ok, heh. Should I submit a PR to remove the hardcoded key?

drzoidberg33:
@erunion Great idea.

JonnyWong16:
The built in Pushover API token has been removed in v1.3.0.

PR #414, opened 2015-12-30
Allow SSL when verifying server in settings and test notifications

JonnyWong16:
Can someone check this to make sure I didn't break anything?

Or maybe there is a better way to do this?
#411, #413, #396


Hellowlol:
Ill check this out when done with the new calendar for htpc manager.

PR #449, opened 2016-01-17
Attempt to start web server on next 5 ports

JonnyWong16:
Good idea?

You won't know what port it's on if you start as a service without console output and without starting the browser. But at least you can get into the webUI to change the port. Easier for new installs?

### DIFF
--- a/PlexPy.py
+++ b/PlexPy.py
@@ -176,7 +176,7 @@ def main():
                         "module to enable HTTPS. HTTPS will be disabled.")
             plexpy.CONFIG.ENABLE_HTTPS = False
 
-    # Try to start the server. Will exit here is address is already in use.
+    # Try to start the server on the next 5 ports. Will exit here is address is already in use.
     web_config = {
         'http_port': http_port,
         'http_host': plexpy.CONFIG.HTTP_HOST,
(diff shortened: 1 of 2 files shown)


roelbeun:
Then you expect the browser opens on install? How do you know the new Port to connect to the webgui then?

boopzz:
Could you do a simple netstat and see if the default port is open. if it is increment port and test again until theres an open one? I suppose you will still have the same prob of it not telling you which port its on. Prob trying to over-engineer it!

JonnyWong16:
That's basically what this does. It checks the default port (8181) if it's in use, then it tries the next one (8182), etc.

boopzz:
I should've just looked at the commit! :smiley: 

`while not web_start and web_config['http_port'] < http_port + 5`

JigSawFr:
This will fail with Docker (i'm using in it), as Docker bind host port (any you want) to docker port. Need possibility to disable this feature by an option in command line. Thanks you !

fliiiix:
:-1: I personally don't like it - but this is just my opinion

jkaberg:
This is a bad solution tbh. The one setting up PlexPy should know what he's doing

PlexPy.py
```diff
@@ -188,7 +188,16 @@ def main():
...
     }
-    webstart.initialize(web_config)
+    web_start = False
+    while not web_start and web_config['http_port'] < http_port + 5:
+        web_start = webstart.initialize(web_config)
+        web_config['http_port'] += 1
+
+    if web_start:
+        http_port = web_config['http_port']
+        plexpy.CONFIG.HTTP_PORT = http_port
```
toddrob99, line 198: If you do go through with this feature, you should log a (warn, error?) here so the user can check the log to see the port it's running on.

PR #461, opened 2016-01-22
Use os.execv instead of subprocess.Popen to restart the process

### LINKED ISSUE #460: Restarting PlexPy via web interface causes it to "fork"
Restarting PlexPy via the web interface causes it to "fork", leaving you with PlexPy running in the background and still printing out to the same stdout.

Here's the steps to reproduce it:
1. Start PlexPy using the following command:
   
   ```
   python2.7 PlexPy.py && echo "PlexPy has exited"
   ```
2. Restart PlexPy via the web interface
3. "PlexPy has exited" will be printed out to the console, as control is returned to the shell. However, PlexPy is still running and any messages are still printed to the console

It looks like we should [`wait`](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait) for the `Popen` call to finish before exiting, as to ensure that the child processes do not "outlive" the parent one.

Adding `.wait()` to the following line in [`plexpy/__init__.py`](https://github.com/drzoidberg33/plexpy/blob/bbf774379d10d1cc9380d3f1fb7fca87a0d0b930/plexpy/__init__.py#L716) seems to fix it:

``` python
subprocess.Popen(popen_list, cwd=os.getcwd()) # Causes weird "forking" effect
```

``` python
subprocess.Popen(popen_list, cwd=os.getcwd()).wait() # No longer causes weird "forking" effect, as we wait for the child to exit before we do
```

A better fix for this would be to replace the `Popen` entirely with a call to [`os.execv`](https://docs.python.org/2/library/os.html#os.execv) which is designed to "replace the current process", preserving all I/O streams and keeping it in the foreground (no weird "forking" effect).

jackwilsdon:
This change fixes #460.

### DIFF
--- a/plexpy/__init__.py
+++ b/plexpy/__init__.py
@@ -708,12 +708,13 @@ def shutdown(restart=False, update=False):
 
     if restart:
         logger.info('PlexPy is restarting...')
-        popen_list = [sys.executable, FULL_PATH]
-        popen_list += ARGS
-        if '--nolaunch' not in popen_list:
-            popen_list += ['--nolaunch']
-        logger.info('Restarting PlexPy with %s', popen_list)
-        subprocess.Popen(popen_list, cwd=os.getcwd())
+        exe = sys.executable
+        args = [exe, FULL_PATH]
+        args += ARGS
+        if '--nolaunch' not in args:
+            args += ['--nolaunch']
+        logger.info('Restarting PlexPy with %s %s', exe, ' '.join(args))
+        os.execv(exe, args)
 
     os._exit(0)
 


JonnyWong16:
Oops,  I commented on the commit instead of the PR. 

Please submit all PRs to the Dev branch, thanks.

jackwilsdon:
@JonnyWong16 It looks like we need to pass both the executable normally **and** in the argument list, as the argument list is used as `argv` by the next process, which always expects `argv[0]` to be the binary itself.

I found [this](http://stackoverflow.com/questions/27460270/how-to-start-a-python-program-recursive-with-os-execv#comment51888952_27460321) on a StackOverflow answer:

> `os.execv` passes the arguments to the executable, and the first argument must be the name of the executable. Many UNIX tools rely on that to alter behaviour based on what symlink was used to execute the tool, for example. 

After doing some testing, it looks like all of the arguments in the `args` array are `str` objects, so we shouldn't have any issues joining them.

I'll re-create the PR on the `dev` branch.

PR #468, opened 2016-01-24
Added {progress_time} parameter (display progress in time format) to notifications

lashkari:
I much prefer to see progress displayed in human readable format (HH:MM:SS) over just the number of minutes or a percentage. This PR adds a new notification substitution parameter called `{progress_time}` that does exactly that.


Hellowlol:
Imo this should use the time format that the user can choose in the settings

lashkari:
> Imo this should use the time format that the user can choose in the settings

Hmm, I didn't really think about that since most of the options in that setting aren't relevant in this context. But I can see how you might want to still control the formatting.

I guess I could first strip out any of the irrelevant substitutions, and then use what's left as the formatting string. So basically, I'd strip out all of these:

```
YYYY    Numeric, 4 digits   Eg., 1999, 2003
YY  Numeric, 2 digits   Eg., 99, 03
MM  Numeric, with leading zeros 01-12
M   Numeric, without leading zeros  1-12
MMMM    Textual full    January-December
MMM Textual three letters   Jan-Dec
DD  Numeric, with leading zeros 01-31
D   Numeric, without leading zeros  1-31
Do  The English suffix for the day of the month st, nd or th in the 1st, 2nd or 15th.
a   am/pm Lowercase am, pm
A   AM/PM Uppercase AM, PM
h   Hour, 12-hour, without leading zeros    1-12
hh  Hour, 12-hour, with leading zeros   01-12
zz  Timezone abbreviation   Eg., EST, MDT ...
```

Thoughts?

JonnyWong16:
I think you can assume that the user hasn't put any year, month, or day, substitutions in the "Time Format" settings. So we just need to strip out the am/pm and timezone.

The formatting can be done using the `arrow` library I added in the last update, so we can use the javascript substitutions instead of the python ones.

See: https://github.com/drzoidberg33/plexpy/blob/master/plexpy/notification_handler.py#L529

lashkari:
Okay, updated to use the time format defined in settings. Thanks for the direction.

JonnyWong16:
I messed up a clean merge. I'm can add this manually, as well as `{stream_time}` and `{remaining_time}`, unless you want to rebase and add those two.

You can use `duration_format` for the formatting.
https://github.com/drzoidberg33/plexpy/blob/dev/plexpy/notification_handler.py#L346

JonnyWong16:
I added these options here 14a90d84ec62756f764e12eeeda3f943c5a448ac

PR #501, opened 2016-01-31
Watched notifications using websockets fix

drzoidberg33:
Annoyingly I am getting a lot of duplicate watched notifications as it appears the state is not yet written to the db before the next one fires off.

I've tried to address this here (as well as fix a bug on the welcome page).

### DIFF
--- a/data/interfaces/default/welcome.html
+++ b/data/interfaces/default/welcome.html
@@ -378,8 +378,8 @@ from plexpy import common
             var pms_ip = $("#pms_ip").val().trim();
             var pms_port = $("#pms_port").val().trim();
             var pms_identifier = $("#pms_identifier").val();
-            var pms_ssl = $("#pms_ssl").val();
-            var pms_is_remote = $("#pms_is_remote").val();
+            var pms_ssl = $("#pms_ssl").is(':checked') ? 1 : 0;
+            var pms_is_remote = $("#pms_is_remote").is(':checked') ? 1 : 0;
             if ((pms_ip !== '') || (pms_port !== '')) {
                 $("#pms-verify-status").html('<i class="fa fa-refresh fa-spin"></i> Validating server...');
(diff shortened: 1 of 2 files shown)


Hellowlol:
When was the ssl and remote wrong? Was this via discover or when entering it manually?

drzoidberg33:
Generally only when setting manually as it looks like the selectize method doesn't use this function.

JonnyWong16:
Changes look ok to me.

drzoidberg33:
Thanks, I'm just trying to still figure out how to not call `notify` at all if there are no `on_watched` triggers. Any ideas?

JonnyWong16:
The only way I can think of is to check through all the `available_notification_agents`.

```
if any(agent['on_watched'] for agent in notifiers.available_notification_agents)
```

drzoidberg33:
Thanks Jonny, I'll get back to this when I have a moment.

drzoidberg33:
Hey @JonnyWong16 want to take a quick peek here, this look okay?

JonnyWong16:
I'm wondering if just removing the new thread on sending the notification will solve the problem without the need for all the if statements. It looks like all websocket events are processed on the same thread.

I feel like this just (unnecessarily) checks the notify state twice. Once here, then again in `notification_handler` (https://github.com/drzoidberg33/plexpy/blob/dev/plexpy/notification_handler.py#L113)

drzoidberg33:
Yes, it should fix the issue but I don't like the fact that `notify` is getting called on every event. it's having to do much more work than just checking if the notify state is set.

JonnyWong16:
Then we change all the `notify` calls to check if the notification state is enabled before the call. Then the check isn't needed in `notification_handler`?

JonnyWong16:
Merged in #586.

Issue #483, opened 2016-01-24
Plexpy hanging after update


realvz:
My plexpy install on 2 different machines is completely hosed. I performed the update manually and after install plexpy starts normally but after homepage is rendered it hangs. I can't even curl the homepage, it results in connection timed out. 

Logs. Update occurred after 4pm on 24 Jan. 
http://paste.ubuntu.com/14655673b

realvz:
In the meanwhile I rolled back to 45c2f5 and things are up again on one machine, the other machine is still running latest update and hanging at startup.

JonnyWong16:
What do you mean by "hangs"? It's just unresponsive and you can't click on any links?

realvz:
The page never finished loading. If I try to refresh it just times out. The page only loads once when I restart the service and that too not completely.  After that all I get is connection timed out.

JonnyWong16:
See if v1.3.2 fixes it. If not, try setting `update_section_ids=0` in the `config.ini` file and starting it up again to see if the library updating is the problem.

erodewald:
@JonnyWong16 I had this issue after doing an update from git about 2 hrs ago (Ubuntu 15.10). The pid was hung/stale so I set `update_section_ids=0` (it was `-1`) and then restarted the service. It seems to be working now.  

edit: Library statistics is empty - I see another thread on that so I'll check that out.

JonnyWong16:
Yeah, `update_section_ids=0` disables the library update (so the stats will be empty).
`update_section_ids=1` starts the update (-1 means its updating).

But it seems like PlexPy just freezes when it's trying to update. And it seems like it only happens on Ubuntu.

erodewald:
@JonnyWong16 I see, I'll try setting it to 1 and see if it hangs again. 

Not hanging - but no stats yet. Possibly still generating them? I'll give it some time before investigating a problem that may not exist.

realvz:
My update_section_ids is already set to 0 and already running latest version, still no dice. Btw it's happening on both Ubuntu 15.10 and raspbian wheezy.

alternativesurfer:
1.3.2 just broke this for me.
Timing out when attempting to load the app.

Update_section_ids is already set to 0

this is on Win server 2012 R2

JonnyWong16:
Everyone post your logs please.

realvz:
Here are my logs. I deleted the log file before restarting, these logs include everything after restarting plexpy service. The web interface never opens for me anymore. 

[...]

PR #502, opened 2016-01-31
Run startup tasks after daemonizing

JonnyWong16:
Any negative effects from moving this?

Possibly solves #483

### DIFF
--- a/PlexPy.py
+++ b/PlexPy.py
@@ -153,11 +153,11 @@ def main():
     # Put the database in the DATA_DIR
     plexpy.DB_FILE = os.path.join(plexpy.DATA_DIR, 'plexpy.db')
 
-    # Read config and start logging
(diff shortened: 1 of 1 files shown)


Hellowlol:
I would move this https://github.com/Hellowlol/plexpy/blob/master/PlexPy.py#L160 above the entire initialize.

Hellowlol:
Let just hope it works :P

JonnyWong16:
How much more can I break it? amirite?

JonnyWong16:
Nope, now the logger is started after the daemon...

Hellowlol:
Just slap it in a print statements. Its just one line and it s warming. I never checked if it would work...

JonnyWong16:
Never mind, init logger on line 95.

Merging, hopefully it doesn't break anything.

PR #540, opened 2016-02-13
SSL certificate generation

an3k:


plexpy/helpers.py
```diff
@@ -341,31 +341,36 @@ def split_string(mystring, splitvar=','):
 
 def create_https_certificates(ssl_cert, ssl_key):
     """
-    Create a pair of self-signed HTTPS certificares and store in them in
+    Create a self-signed HTTPS certificate and store it in
     'ssl_cert' and 'ssl_key'. Method assumes pyOpenSSL is installed.
 
-    This code is stolen from SickBeard (http://github.com/midgetspy/Sick-Beard).
+    The code were noted was stolen from SickBeard (http://github.com/midgetspy/Sick-Beard).
```
Arcanemagus, line 347: The previous version of this sentence was much better, why did you change it?

JonnyWong16:
#541.

plexpy/helpers.py
an3k, line 347: Most of the code wasn't even from SickBeard anymore, but you're right that my version sounded a bit bad.

PR #541, opened 2016-02-13
Create self-signed HTTPS certificates

JonnyWong16:
@drzoidberg33 @Hellowlol 

Want to double check this?


Arcanemagus:
Total cert noob here: Would it be possible to get _real_ certs through https://letsencrypt.org/?

Hellowlol:
Ill test it later today.

Hellowlol:
@Arcanemagus Yes it would, but i think that is something they have to setup them selfes since they are only valid in 60 days (renewal is trivial and a script can run with cron). This pr is about getting a simple self signed cert. If a use wants to use a real one its easy to add in the settings.

Hellowlol:
Worked fine! Nice work.

PR #548, opened 2016-02-16
Add all current api calls to the docs

Hellowlol:
This document has been created from docs_md command. Its just to list whats currently available.
The docs are as good as plexpy own docs :(
I think it would be wise to switch to google style docs in the entire project as its much simpler to parse to something decent.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,4 +1,5 @@
 # API Reference
+
 The API is still pretty new and needs some serious cleaning up on the backend but should be reasonably functional. There are no error codes yet.
 
 ## General structure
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -23,58 +24,386 @@ Response example
...
+Get all details about a user from plextv
+
+
+### get_user_list
+
+
+### notify
+
+
+### random_arnold_quotes
```
drzoidberg33, line 376: The only good reason to have this api. :laughing:

Hellowlol, line 376: Well when all you need to do is slap decorator to add it to the api this is what happens... :P, while that being said, some stuff should be imported manually. I think i went a little overboard..

JonnyWong16:
I want to go through and add documentation for everything first.

Hellowlol:
Alright

PR #575, opened 2016-02-24
Few improvements to the FreeNAS init script

chiviak:
This PR modifies the following:
1. Makes the FreeNAS init script executable
2. Calls PlexPy directly using daemon mode instead of wrapping it in the daemon program
3. Allows for a 'plexpy_flags' argument so the caller can specify custom flags like '--port'


JonnyWong16:
Can you resubmit this PR to the dev branch? And maybe update the FreeBSD init-script to match?

PR #594, opened 2016-03-06
Few improvements to the FreeNAS/FreeBSD init scripts

chiviak:
This PR modifies the following:
1. Makes the FreeNAS init script executable
2. Calls PlexPy directly using daemon mode instead of wrapping it in the daemon program
3. Allows for a 'plexpy_flags' argument so the caller can specify custom flags like '--port'


JonnyWong16:
Thanks.

Arcanemagus:
@chiviak Did you actually _test_ this on FreeNAS?

This completely breaks inside the jail I have setup for PlexPy as this then relies on the `#!/usr/bin/env python` inside PlexPy.py to work properly... which it doesn't of course because `python` is `python2` on that platform.

``` ShellSession
root@plexpy:/usr/local/share/plexpy # service plexpy start
Starting plexpy.
env: python: No such file or directory
/usr/local/etc/rc.d/plexpy: WARNING: failed to start plexpy
```

chiviak:
@Arcanemagus Thanks for letting me know. I definitely tested the changes, I wouldn't have submitted them if I wasn't using them myself. I use FreeNAS religiously and used the changes for about a week before submitting the PR. 

As far as I recall, a symlink from python2 to python should have be created by the python package during installation so I'm not sure why it doesn't exist on your system, different jail versions perhaps. Here is what I get when looking for python on my Plex jail (installed via the plugin):

```
root@plexmediaserver_1:/ # which python                                                                                             
/usr/local/bin/python                                                                                                               
root@plexmediaserver_1:/ # ll /usr/local/bin/python*                                                                                
lrwxr-xr-x  1 root  wheel     7 Sep  2  2015 /usr/local/bin/python@ -> python2                                                      
lrwxr-xr-x  1 root  wheel    14 Sep  2  2015 /usr/local/bin/python-config@ -> python2-config                                        
lrwxr-xr-x  1 root  wheel     9 Sep  2  2015 /usr/local/bin/python2@ -> python2.7                                                   
lrwxr-xr-x  1 root  wheel    16 Sep  2  2015 /usr/local/bin/python2-config@ -> python2.7-config                                     
-r-xr-xr-x  1 root  wheel  5176 Sep  2  2015 /usr/local/bin/python2.7*                                                              
-r-xr-xr-x  1 root  wheel  1687 Sep  2  2015 /usr/local/bin/python2.7-config*
```

The quick fix would be to add the symlink manually on your system by doing the following:

```
ln -s /usr/local/bin/python2 /usr/local/bin/python
```

Though if you prefer to not create a symlink manually then let's work together to find a solution which works for everyone.  Off the top of my head we can add the following:

```
# Check for the python version before executing PlexPy                                                                              
plexpy_command="${plexpy_dir}/PlexPy.py"                                                                                            
plexpy_initial_args="--daemon"                                                                                                      
which python > /dev/null 2>&1                                                                                                      
if [ ! $? -eq 0 ]; then                                                                                                             
  which python2 > /dev/null 2>&1                                                                                                    
  if [ $? -eq 0 ]; then                                                                                                             
    plexpy_command="/usr/sbin/daemon"                                                                                               
    plexpy_initial_args="`which python2` ${plexpy_dir}/PlexPy.py"                                                                   
  else                                                                                                                              
    echo "Oops, you must have either python or python2 installed!"                                                                  
    exit 1                                                                                                                          
  fi                                                                                                                                
fi                                                                                                                                  

command="${plexpy_command}"                                                                                                         
command_args="${plexpy_initial_args} --pidfile ${plexpy_pid} --quiet --nolaunch ${plexpy_flags}"
```

What are your thoughts?

JonnyWong16:
I tested this too and mine is symlinked as well. Of course that still doesn't mean everyone's jails work the same way.

Arcanemagus:
That was a great idea about differing jail versions:
- `pluginjail` type jails do **not** have the `python` symlink
- `VirtualBox-4.3.12` type jails do **not** have the `python` symlink (PlexPy is apparently currently based in this)
- `FreeNAS 9.3 pluginjail` type jails **do** have the symlink

So it looks like newer jails have the symlink, but all older jails do not. I'm honestly not sure which way I think this should be fixed... most people probably aren't going to have ancient jails lying around, and newer jails will have it.

Do we do a workaround or just leave it to users to fix this if they are running in an older jail?

chiviak:
@Arcanemagus , ultimately I think it's up to @JonnyWong16 to decide which solution to go with since he will most likely be the one fielding questions if anyone has an issue in the future.

I've submitted PR https://github.com/drzoidberg33/plexpy/pull/607 which contains the script changes I illustrated above, they seem to be working for me but it'd be great if you could give it a test too and make sure it works on your environment. If we decide to not bother putting them in that's fine too, I just wanted to make sure they were available. I'm open to additional ideas as well if there's a cleaner solution.

PR #607, opened 2016-03-14
Change FreeNAS start command based on available versions of python

chiviak:
This will check for the available version of python and execute a different start command depending on what is found.

See comments on PR https://github.com/drzoidberg33/plexpy/pull/594 for more context.

### DIFF
--- a/init-scripts/init.freenas
+++ b/init-scripts/init.freenas
@@ -38,8 +38,23 @@ load_rc_config ${name}
 status_cmd="${name}_status"
 stop_cmd="${name}_stop"
 
-command="${plexpy_dir}/PlexPy.py"
-command_args="--daemon --pidfile ${plexpy_pid} --quiet --nolaunch ${plexpy_flags}"
+# Check for the python version before executing PlexPy
+plexpy_command="${plexpy_dir}/PlexPy.py"
+plexpy_initial_args="--daemon"
+which python > /dev/null 2>&1
+if [ ! $? -eq 0 ]; then
+  which python2 > /dev/null 2>&1
+  if [ $? -eq 0 ]; then
+    plexpy_command="/usr/sbin/daemon"
+    plexpy_initial_args="`which python2` ${plexpy_dir}/PlexPy.py"
+  else
+    echo "Oops, you must have either python or python2 installed!"
+    exit 1
+  fi
+fi
+
+command="${plexpy_command}"
+command_args="${plexpy_initial_args} --pidfile ${plexpy_pid} --quiet --nolaunch ${plexpy_flags}"
 
 # Ensure user is root when running this script.
 if [ `id -u` != "0" ]; then


JonnyWong16:
I probably have no idea what I'm doing, but what if we just use:

```
command="/usr/local/bin/python2"
command_args="${plexpy_dir}/PlexPy.py --daemon --pidfile ${plexpy_pid} --quiet --nolaunch ${plexpy_flags}"
```

It seems like all FreeNAS jails will have `python2`.

chiviak:
I thought about that as well and actually tried it too.  The way you have it written out wouldn't actually work due to the way that services are invoked. Although that works on command line, through a service the 'command' and 'command_args' aren't just concatenated and executed, the 'command_args' are actually passed in to the 'command' directly. So to attempt it, it'd have to be the following:

```
command="/usr/local/bin/python2 ${plexpy_dir}/PlexPy.py"                                                                            
command_args="--daemon --pidfile ${plexpy_pid} --quiet --nolaunch ${plexpy_flags}"
```

I initially thought this would work but when I attempted it, I got the following error:

```
root@plexmediaserver_1:/usr/pbi/plexpy # service plexpy start                                                                       
/usr/local/etc/rc.d/plexpy: WARNING: no shebang line in /usr/local/bin/python2                                                      
/usr/local/etc/rc.d/plexpy: WARNING: run_rc_command: cannot run /usr/local/bin/python2 /usr/pbi/plexpy/PlexPy.py
```

Which I'm honestly not sure why it's thrown or how to resolve it.  I did some searching but didn't really find any solution.

So I did a bit more reading about rc commands and found a cleaner solution for this. If we make the assumption that all FreeNAS boxes have python2 available then we can actually override the start command (in the same way that status_cmd and stop_cmd are currently done) and use the following:

```
plexpy_start() {                                                                                                                    
    verify_plexpy_pid                                                                                                               
    if [ $? -eq 0 ]; then                                                                                                           
        echo "Starting $name."                                                                                                      
        `which python2` ${plexpy_dir}/PlexPy.py --daemon --pidfile ${plexpy_pid} --quiet --nolaunch ${plexpy_flags}                 
    else                                                                                                                            
        echo "$name is already running."                                                                                            
    fi                                                                                                                              
}
```

I just tested it out and it worked! If we want to take it step further then we can incorporate some of the script changes in this PR to determine the python version first, then simply use that version explicitly for start.

JonnyWong16:
I found [this](https://stackoverflow.com/questions/18993438/shebang-env-preferred-python-version).

Basically add these lines to the top of `PlexPy.py`.

```
#!/bin/sh
''''which python    >/dev/null 2>&1 && exec python    "$0" "$@" # '''
''''which python2   >/dev/null 2>&1 && exec python2   "$0" "$@" # '''
''''which python2.7 >/dev/null 2>&1 && exec python2.7 "$0" "$@" # '''
''''exec echo "Error: Python not found!" # '''
```

Leave the init script as is, and PlexPy will try to run with `python`, `python2`, or `python2.7`.

Anyone see any downsides to this? Other systems?

Arcanemagus:
That seems like it should work on any system with a working `/bin/sh` from the explanation in the linked SO answer. :+1:

(It works perfectly in my problem jail btw.)

JonnyWong16:
Added in c8f1cb0a0af5d912d3afc0fef19091ef4942065a

chiviak:
@JonnyWong16 Thanks!

PR #612, opened 2016-03-16
Add CherryPy Environment Option

codedecay:
Add `http_environment` option allowing user to set CherryPy environment from PlexPy `config.ini`. Default for new option is set to `production`.

CherryPy Built-in environment options: `production`, `staging`, `embedded`, `test_suite`

CherryPy's environment option allows for groups of default configuration settings to be set via a single option. The following shows the option sets for each of the default environments.
##### Source from _cpconfig.py

``` python
Config.environments = environments = {
    "staging": {
        'engine.autoreload_on': False,
        'checker.on': False,
        'tools.log_headers.on': False,
        'request.show_tracebacks': False,
        'request.show_mismatched_params': False,
        },
    "production": {
        'engine.autoreload_on': False,
        'checker.on': False,
        'tools.log_headers.on': False,
        'request.show_tracebacks': False,
        'request.show_mismatched_params': False,
        'log.screen': False,
        },
    "embedded": {
        # For use with CherryPy embedded in another deployment stack.
        'engine.autoreload_on': False,
        'checker.on': False,
        'tools.log_headers.on': False,
        'request.show_tracebacks': False,
        'request.show_mismatched_params': False,
        'log.screen': False,
        'engine.SIGHUP': None,
        'engine.SIGTERM': None,
        },
    "test_suite": {
        'engine.autoreload_on': False,
        'checker.on': False,
        'tools.log_headers.on': False,
        'request.show_tracebacks': True,
        'request.show_mismatched_params': True,
        'log.screen': False,
        },
    }
```

### DIFF
--- a/PlexPy.py
+++ b/PlexPy.py
@@ -181,6 +181,7 @@ def main():
         'http_port': http_port,
         'http_host': plexpy.CONFIG.HTTP_HOST,
         'http_root': plexpy.CONFIG.HTTP_ROOT,
+        'http_environment': plexpy.CONFIG.HTTP_ENVIRONMENT,
         'http_proxy': plexpy.CONFIG.HTTP_PROXY,
         'enable_https': plexpy.CONFIG.ENABLE_HTTPS,
         'https_cert': plexpy.CONFIG.HTTPS_CERT,
--- a/plexpy/config.py
+++ b/plexpy/config.py
@@ -142,6 +142,7 @@ _CONFIG_DEFINITIONS = {
     'HTTPS_KEY': (str, 'General', ''),
     'HTTPS_DOMAIN': (str, 'General', 'localhost'),
     'HTTPS_IP': (str, 'General', '127.0.0.1'),
+    'HTTP_ENVIRONMENT': (str, 'General', 'production'),
     'HTTP_HOST': (str, 'General', '0.0.0.0'),
     'HTTP_PASSWORD': (str, 'General', ''),
     'HTTP_PORT': (int, 'General', 8181),
--- a/plexpy/webstart.py
+++ b/plexpy/webstart.py
@@ -46,6 +46,7 @@ def initialize(options):
     options_dict = {
         'server.socket_port': options['http_port'],
         'server.socket_host': options['http_host'],
+        'environment': options['http_environment'],
         'server.thread_pool': 10,
         'tools.encode.on': True,
         'tools.encode.encoding': 'utf-8',


Hellowlol:
In not sure this will be usefull for plexpy imo a traceback should always showed in the browser and I don't see what this brings to the table.

codedecay:
Tracebacks reveal paths to unauthenticated requests which are a major security concern. If you are running plexpy from a home directory a user can determine a valid username for the server by simply canceling the basic authentication dialog.

Hellowlol:
Your making a lot of assumptions. I do t know what other people do but I don't use a home folder or my name as username, even if they got it they still have to 16 chars password. I think I'm good ;)

codedecay:
That is one example for how it can leak important information but in general any path disclosure is a security concern. The CherryPy authors disable tracebacks in the 'production' environment by default and imo PlexPy should as well. Many users of this product might not have a hardened server or strong passwords. You should side with protecting users from themselves by default by making the default environment 'production'.

[OWASP Full Path Disclosure](https://www.owasp.org/index.php/Full_Path_Disclosure)

JonnyWong16:
I assume this will be disabled in the browser, but errors will still be available in the logger.

codedecay:
Correct, setting the show_traceback option to false only prevents tracebacks from being included in the response body.

##### Source from _cprequest.py

``` python
    show_tracebacks = True
    """
    If True, unexpected errors encountered during request processing will
    include a traceback in the response body."""

    show_mismatched_params = True
    """
    If True, mismatched parameters encountered during PageHandler invocation
    processing will be included in the response body."""
```

JonnyWong16:
I tested and tracebacks are not sent to the logger.

They need to be sent to the logger: https://cherrypy.readthedocs.org/en/3.3.0/refman/_cplogging.html#custom-handlers

codedecay:
Can you post your unit test so I can work off of it. I'll write a custom handler if that is necessary to log tracebacks.

JonnyWong16:
I just did something like `/home?abc=123` to get the error `Unexpected query string parameters: abc`

I also added something like this:

```
for h in logger.logger.handlers:
    cherrypy.log.error_log.addHandler(h)
```

to `webstart.py`.

Hellowlol:
Is there a option there the reloaded is activated? All the options seem to do the same thing.

JonnyWong16:
I assume `engine.autoreload_on: True` is the setting, but no idea how it works.

Hellowlol:
Ye, I want to add a dev option that turns off the blacklist filter, force debug log, stops the browser open and activates the reloaded like htpc manager has.

JonnyWong16:
I think I got autoreload figured out....

JonnyWong16:
Merging so I can build off of it.

PR #629, opened 2016-03-26
Fix unicode logging with 1252 encoded locale on Windows

alshain:
On my system (Windows, Swiss-German locale) I get loads of UnicodeDecodeErrors instead of log messages.

Apparently, my locale makes "%b" evaluate to "Mär". Combining this with unicode log messages makes the logger throw an exception during formatting. A simple workaround that I've found is to replace the abbreviation with a numeric month representation.

Changing the format strings to be unicode themselves does not help, unfortunately. I suspect it's connected to Windows' locale encoding being Windows-1252.

### DIFF
--- a/plexpy/logger.py
+++ b/plexpy/logger.py
@@ -160,7 +160,7 @@ def initLogger(console=False, log_dir=False, verbose=False):
     if log_dir:
         filename = os.path.join(log_dir, FILENAME)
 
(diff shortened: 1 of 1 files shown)


JonnyWong16:
I'm all for using ISO date formats. Thanks.

JonnyWong16:
I was a bit hasty.

Please submit PR to the dev branch.

PR #633, opened 2016-03-26
Allow formatting of IFTTT event key with action name

evilmarty:
The existing IFTTT implementation is very limiting as to what you do with the service because all events emitted use the same name and IFTTT does offer any mechanism to filter actions based on the subject. This change allows for enriching the event name sent to Maker thus allowing better control of actions to be carried out. ie. Turn lights off when playing a movie and turn them back on when stopped. 

To avoid breaking existing implementations the feature is op-in by including a `{action}` placeholder in the event name. ie. `plexpy_{action}`.

### DIFF
--- a/plexpy/notifiers.py
+++ b/plexpy/notifiers.py
@@ -436,7 +436,7 @@ def get_notification_agent_config(agent_id):
         return []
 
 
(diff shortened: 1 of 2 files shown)


JonnyWong16:
Please submit pull requests to the dev branch.

PR #634, opened 2016-03-26
Allow formatting of IFTTT event key with action name

evilmarty:
The existing IFTTT implementation is very limiting as to what you do with the service because all events emitted use the same name and IFTTT does offer any mechanism to filter actions based on the subject. This change allows for enriching the event name sent to Maker thus allowing better control of actions to be carried out. ie. Turn lights off when playing a movie and turn them back on when stopped. 

To avoid breaking existing implementations the feature is op-in by including a `{action}` placeholder in the event name. ie. `plexpy_{action}`.

### DIFF
--- a/plexpy/notifiers.py
+++ b/plexpy/notifiers.py
@@ -436,7 +436,7 @@ def get_notification_agent_config(agent_id):
         return []
 
 
-def send_notification(agent_id, subject, body, **kwargs):
+def send_notification(agent_id, subject, body, notify_action, **kwargs):
     if str(agent_id).isdigit():
         agent_id = int(agent_id)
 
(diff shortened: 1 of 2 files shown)


JonnyWong16:
You can also update the description [here](https://github.com/drzoidberg33/plexpy/pull/634/files#diff-e1b997e1131edcd9a37a7578818b5862R1653) to indicate that you can use `{action}` in the event name.

Also,

```
git pull --rebase origin dev
```

to no include all the merge commits.

evilmarty:
Updated the description and rebased with `dev`.

JonnyWong16:
Works, thanks!

Issue #382, opened 2015-12-19
Graph data per user


DFKnight:
I'm not sure if this is already a feature in PlexPy, but I would like to be able to view the data shown in the Graphs page on a per-user basis.

BehrozKarimi:
+1

JonnyWong16:
Added in v1.4.0.

PR #662, opened 2016-04-09
Added ability to view per-user graphs

otgerrogla:
Added a `<select>` box at the top of the Graphs page, which allows selecting a single user to see only their data in the graphs. By default it has "All users" selected, which works as before showing the overall statistics.

Implements the feature requested in issue #382


JonnyWong16:
I finally got around to testing this. It looks great!

Can you also filter the "Play count by top 10 users" and "Play count by user and stream type" to the selected user? I know it will only show one bar, but it will be consistent with the other graphs.

plexpy/users.py
```diff
@@ -21,6 +21,23 @@ class Users(object):
...
+        try:
+            query = 'SELECT user_id, ' \
+                    '(CASE WHEN users.friendly_name IS NULL OR TRIM(users.friendly_name) = "" \
+                    THEN users.username ELSE users.friendly_name END) AS friendly_name ' \
+                    'FROM users ' \
+                    'WHERE deleted_user = 0'
+
+            result = monitor_db.select(query)
+        except Exception as e:
+            logger.warn(u"PlexPy Graphs :: Unable to execute database query for get_user_names: %s." % e)
```
JonnyWong16, line 36: Typo:

```
PlexPy Users :: Unable to execute database query for get_names
```

otgerrogla:
Done! I added filtering to those two graphs too.

As for the "typo", actually other functions below that one I added in "users.py" also had mismatching names/error messages. I think the messages correspond to the API call names at "webserve.py". Anyway, I changed the function name so that both are the same.

JonnyWong16:
Thanks, looks good.

Merging so I can use your updated SQL queries in #677.

PR #671, opened 2016-04-16
Optimized iOS, Android and IE10 images

alotufo:
Ran iOS, Android and IE10 images through ImageOptim to optimize and reduce file size


JonnyWong16:
You need to submit from dev to dev so you don't get all the merge commits.

PR #672, opened 2016-04-16
Optimized iOS, Android and IE10 images

alotufo:
Ran iOS, Android and IE10 images through ImageOptim to reduce file size


JonnyWong16:
Looks good. Thanks!

Issue #54, opened 2015-08-12
[Create styled HTML login page] Allow password to be saved by browsers


alexandrupaulpopa:
Hi,

Can you please allow password to be saved by browsers, it would make the login in process just a bit easier

THanks

drzoidberg33:
I'm guessing you mean you want a HTML login page? Currently authentication is just using HTTP Basic auth which will use your browsers pop-up to get the username and password.

alexandrupaulpopa:
Thats exactly it. Thanks

alexandrupaulpopa:
any updates on this issue?

JonnyWong16:
I tried. Didn't work. Got confused. Gave up.

If anyone else attempts this, I have the html login template made. :stuck_out_tongue:

drzoidberg33:
Hehe, thanks Jonny :smile:

I'll give it a go when I have some time.

alexandrupaulpopa:
Updates?

drzoidberg33:
Nobody has had time to work on this yet as far as I'm aware.

Hellowlol:
@JonnyWong16 can you send me the login screen in a gist?
@drzoidberg33 
I also have question how we should handle this. 
Should the users be added manually. Picked from the friends list? 
Should the views be limited in any way only admin as access to settings?
Should there be a restricted user limited to only see page xxx?

JonnyWong16:
You''ll have to wait until January. I'm on vacation and I don't have access to my computer.

drzoidberg33:
You and me both :christmas_tree:

Hellowlol:
Np, no rush.

jkaberg:
> Should the users be added manually. Picked from the friends list? 

Friends list, eg. only users invited to browse the server are able to login

> Should the views be limited in any way only admin as access to settings?

Admin/Owner only for settings

> Should there be a restricted user limited to only see page xxx?

Yes, or well I don't want to see users being able to see other users activity etc. Privacy concerns you know. Only the owner should be able to see the activity of all users

my 2 cents :-)

Hellowlol:
1. Easy
2. Easy
3. There is no way i can do this. My sql sucks.

jkaberg:
I don't know about cherrpy but if in flask (with flask-security) you can implement decorators for each view, eg

```
@requires_admin
def setttings_page():
    pass

@login_required
def recent_activity_page():
    pass
```

[...]

PR #676, opened 2016-04-20
Initial implementation of login control

JonnyWong16:
#54 and [FeatHub#1](http://feathub.com/drzoidberg33/plexpy/+1)

I finally got it working!
- Adds an HTML login page
- No longer uses cherrypy basic auth
- Uses the admin account set in Settings > Access Control
- Currently only the main pages (tabs) are protected
- Redirect to login page when trying to access a protected page
  - Redirect back to `from_page` after successful login

@drzoidberg33 @Hellowlol 
I _really_ want you guys to make sure that there are no holes in this before I merge.

Todo list:
- [ ] Protect the remaining URIs (not sure how to handle the api decorator with authentication)
- [ ] Add logout button (endpoint `/auth/logout`)
- [ ] Make "Remember me" checkbox actually do something
- [ ] Login error message when failed attempt
- [ ] Setup user groups for guest access (probably a separate PR)

### DIFF
--- a/data/interfaces/default/base.html
+++ b/data/interfaces/default/base.html
@@ -12,14 +12,14 @@ from plexpy.helpers import anon_url
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta name="description" content="">
     <meta name="author" content="">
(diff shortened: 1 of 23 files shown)


JonnyWong16:
Dammit...dev branch...

PR #677, opened 2016-04-20
Initial implementation of login control

JonnyWong16:
#54 and [FeatHub#1](http://feathub.com/drzoidberg33/plexpy/+1)

I finally got it working!
- Adds an HTML login page
- No longer uses cherrypy basic auth
- Uses the admin account set in Settings > Access Control
- ~~Currently only the main pages (tabs) are protected~~
- Redirect to login page when trying to access a protected page
  - ~~Redirect back to `from_page` after successful login~~ Loads homepage after successful login
- Allows guest login using Plex.tv accounts (if guest access is enabled)

@drzoidberg33 @Hellowlol 
I _really_ want you guys to make sure that there are no holes in this before I merge.

Todo list:
- [x] Fix login redirects when using a base url (`http_root`)
- [x] Hash password in config file (~~use [passlib](https://pythonhosted.org/passlib/lib/passlib.hash.pbkdf2_digest.html#passlib.hash.pbkdf2_sha256) library~~)
- [x] Protect the remaining URIs (not sure how to handle the api decorator with authentication -  protected when accessing directly `/endpoint`, but open with using `/api/v2?cmd=endpoint` with valid api token)
- [x] Add logout button (endpoint `/auth/logout`)
- [x] Make "Remember me" checkbox actually do something (sets session for 30 days)
- [x] Login error message when failed attempt
- [x] Admin login popout from menu or modal
- [x] Setup user groups for guest access ~~(probably a separate PR)~~
  - [x] Update `users` table to include `allow_guest`, `user_token`, `server_token`, ~~`session_id`, and `last_login`~~ (sessions handled by CherryPy)
  - [x] Retrieve a Plex token for each user ~~(need to figure out how to do this like [PlexRequests](https://github.com/lokenx/plexrequests-python/issues/2))~~
    - [x] Retrieve a user access token
    - [x] Retrieve a server access token
    - [x] Save user and server access tokens to database
  - [x] Add global toggle for guest access in Settings > Access Control
  - [x] Update SQL queries to filter based on logged in user (or just list comprehension after the query? Which is faster?)
  - [x] Update Plex API calls to use the guest `server_token` instead of the admin's `pms_token` ~~(not really necessary?)~~
  - [x] Apply `content_rating` and `labels` filter for guests
  - [x] Toggle to allow access per user (Users > Edit mode)
  - [x] Go through all the templates to hide admin stuff (all returned data needs to be filtered at the server level not the template level)
    - [x] Home (should we show all current activity/stats with masked user info?)
    - [x] Libraries + Individual library page + Info page (only show shared libraries)
    - [x] Users + Individual user page
    - [x] History
    - [x] Graphs
    - [x] Synced Items
    - [x] Settings (disabled for guests)
    - [x] Info


Hellowlol:
I'll test it this weekend.

Hellowlol:
You need to handle the webdir when you redirect.

JonnyWong16:
What do you mean?

Hellowlol:
If you redirect to / and the user has set a webdir cherrypy will barf :P

JonnyWong16:
Oh you mean the `http_root` (like for reverse proxies). I thought that should work already, but I didn't test it.

Hellowlol:
I didnt really test it, but i experianced that with htpc manager so i just assumed it will happen :P

JonnyWong16:
Fixed the redirect problem.

Hellowlol:
i tested it and it seems to to work just fine. It works with the api decorator aswell atleast the two functions i tested.

JonnyWong16:
Ok, I just want to make sure there are not loopholes around the login screen.

JonnyWong16:
Test again? All admin login should be working now.

Hellowlol:
getLog needs to be protected and i dont think auth should be used for the static resources and the redirect from_page needs to be fixed. The rest looks good (i havnt tested everything :P)

JonnyWong16:
Ah, I missed `getLogs`. Good catch. There is no `from_page` anymore, it just redirects to home.

I can access the static resources (images, css, js) without logging in?

plexpy/users.py
```diff
@@ -15,6 +15,63 @@
 
 from plexpy import logger, datatables, common, database, helpers
 
+def user_login(username=None, password=None):
```
JonnyWong16: Someone please check my logic in this function.

drzoidberg33:
Pulled this branch in and getting 500 Server errors. You still in-between things?

JonnyWong16:
Not sure what's going on there...It's working for me.

drzoidberg33:
> Not sure what's going on there...It's working for me.

Hmm. You starting with a fresh install?

capture

JonnyWong16:
Ah, I haven't tested a new install. 

Started with the `--dev` flag and found the problem. Will fix when I get home.

Hellowlol:
@JonnyWong16 Yes, just turn off sessions and auth for the static paths. You can see https://github.com/Hellowlol/HTPC-Manager/blob/master2/htpc/server.py#L135 for a example.

plexpy/helpers.py
```diff
@@ -562,4 +562,18 @@ def uploadToImgur(imgPath, imgTitle=''):
...
+
+def filter_datatable_session(list_of_dicts):
+    import cherrypy
+    from plexpy.webauth import SESSION_KEY
+
+    _cp_session = cherrypy.session.get(SESSION_KEY)
+    _session = {}
+    _session['username'], _session['user_id'], _session['user_group'], _session['expiry'] = \
+        _cp_session if _cp_session else (None, None, None, None)
+
```
Hellowlol: Move str(_session['user_id']) outside the loop.

plexpy/__init__.py
```diff
@@ -740,6 +743,21 @@ def dbcheck():
             'ALTER TABLE users ADD COLUMN deleted_user INTEGER DEFAULT 0'
         )
 
+    # Upgrade users table from earlier versions
+    try:
+        c_db.execute('SELECT allow_guest FROM users')
+    except sqlite3.OperationalError:
+        logger.debug(u"Altering database. Updating database table users.")
+        c_db.execute(
+            'ALTER TABLE users ADD COLUMN allow_guest INTEGER DEFAULT 1'
```
Hellowlol: Id prefer it guest access was disabled by default.

plexpy/config.py
```diff
@@ -132,23 +133,26 @@ def bool_int(value):
...
-    'HOME_STATS_CARDS': (list, 'General', ['top_tv', 'popular_tv', 'top_movies', 'popular_movies', 'top_music', \
+    'HOME_STATS_CARDS': (list, 'General', ['top_movies', 'popular_movies', 'top_tv', 'popular_tv', 'top_music', \
         'popular_music', 'last_watched', 'top_users', 'top_platforms', 'most_concurrent']),
     'HTTPS_CREATE_CERT': (int, 'General', 1),
     'HTTPS_CERT': (str, 'General', ''),
     'HTTPS_KEY': (str, 'General', ''),
     'HTTPS_DOMAIN': (str, 'General', 'localhost'),
     'HTTPS_IP': (str, 'General', '127.0.0.1'),
     'HTTP_ENVIRONMENT': (str, 'General', 'production'),
+    'HTTP_HASH_PASSWORD': (int, 'General', 0),
```
Hellowlol, line 149: Why are there two off this?

JonnyWong16, line 149: `http_hash_password` is to tell if I need to hash a new password.
`http_hashed_password` is to tell if the password in the config file is already hashed.

plexpy/users.py
```diff
@@ -15,29 +15,75 @@
...
+            return True
+
+        # Otherwise it is a new user or token is no longer valid.
+        # Check if the user is in the database, not deleted, and 'allow_guest' access.
+        user_details = user_data.get_details(user_id=user_id)
+        if user_id == str(user_details['user_id']) and \
+            not user_details['deleted_user'] and user_details['allow_guest']:
+
+            # The user is in the database, so try to retrieve a new server token.
+            # If a server token is returned, then the user is a vaild friend
```
Hellowlol: typo valid

JonnyWong16:
@drzoidberg33 It should work on a new install now.

plexpy/webauth.py
```diff
@@ -0,0 +1,179 @@
...
+        username, user_id, user_group, expiry = cp_sesssion if cp_sesssion else (None, None, None, None)
+
+        if (username and expiry) and expiry > datetime.now():
+            cherrypy.request.login = username
+            for condition in conditions:
+                # A condition is just a callable that returns true or false
+                if not condition():
+                    raise cherrypy.HTTPRedirect(plexpy.HTTP_ROOT)
+        else:
+            raise cherrypy.HTTPRedirect("auth/login")
```
Hellowlol: Does this work with a HTTP_ROOT?

JonnyWong16: From my testing it does.

Hellowlol:
Guest logins can see everything on now playing, dunno if thats intentional or not.

Hellowlol:
Alright, looks good really nice work :)

JonnyWong16:
Yeah, I don't know if we should show all current activity/stats with masked user info? Or only show the logged in user?

Also libraries need to be filtered to show only the shared libraries for the logged in user.

You can also reach certain pages manually. (i.e `/user?user_id=xxxxx`) Haven't decided on how to filter these yet.

Hellowlol:
I would have showed it masked. I also liked that you added a donate button, i hope you get some!

Arcanemagus:
A vote for masked here as well, I would love to be able to let everyone access this once this is implemented so they can see what the current load on the server looks like.

JonnyWong16:
Masking on the homepage is working so far.
- Any activity not by the logged in user shows up as "Plex User"
- If the activity is from a library not shared with the logged in user, then the metadata is also masked.
  - e.g. Movies is shared with User A, but User B is playing a TV Show, the the title/artwork/etc. won't show up.

Same goes for the Watch Statistics and Library Statistics.

JonnyWong16:
I broke background processes...nothing runs unless someone is logged in...

Help? Anyone know a better way to exclude cherrypy threads than ff01c1a?

drzoidberg33:
Getting error on Now Playing when something is playing from a library section the user doesn't have access to:

```
Error !
KeyError: 'year'
% endif </div> <div class="dashboard-activity-metadata-subtitle"> % if a['media_index'] or a['parent_media_index'] or a['parent_title'] or a['title'] or a['year']: % if a['media_type'] == 'episode': <span title="S${a['parent_media_index']} &middot; E${a['media_index']}">S${a['parent_media_index']} &middot; E${a['media_index']}</span> % elif a['media_type'] == 'movie': <span title="${a['year']}">${a['year']}</span> % elif a['media_type'] == 'track':
C:\Users\timvd\Development\Personal\PlexPy\Forks\plexpy\data/interfaces/default/current_activity.html, line 267:
% if a['media_index'] or a['parent_media_index'] or a['parent_title'] or a['title'] or a['year']:
C:/Users/timvd/Development/Personal/PlexPy/Forks/plexpy\lib/mako\runtime.py, line 860:
callable_(context, *args, **kwargs)
C:/Users/timvd/Development/Personal/PlexPy/Forks/plexpy\lib/mako\runtime.py, line 835:
_exec_template(inherit, lclcontext, args=args, kwargs=kwargs)
C:/Users/timvd/Development/Personal/PlexPy/Forks/plexpy\lib/mako\runtime.py, line 803:
**_kwargs_for_callable(callable_, data))
C:/Users/timvd/Development/Personal/PlexPy/Forks/plexpy\lib/mako\template.py, line 443:
return runtime._render(self, self.callable_, args, data)
C:\Users\timvd\Development\Personal\PlexPy\Forks\plexpy\plexpy\webserve.py, line 49:
_session=_session, **kwargs)
```

Hellowlol:
@JonnyWong16 Where does it break background processing? I dont get it. Please try to keep the pr on point (not do libs update etc in it) as it monster huge is getting hard to review :P

JonnyWong16:
`sessions.py` L30. If there is no session then `cherrypy.session.get[SESSION_KEY]` returns `None`. 

I "fixed" it by excluding cherrypy threads from any kind of session filtering/masking.

Hellowlol:
Okei, well if cherrypy.sesssion dont not throw a exception you could do.

```
x =       {'user_id': None,
            'user': None,
            'user_group': 'admin',
            'user_libraries': None,
            'expiry': None}

return cherrypy.session.get(SESSION_KEY, x)
```

JonnyWong16:
Hmm right. 

@drzoidberg33 was the current activity a track? I think it's because tracks are missing the `year` key here
https://github.com/JonnyWong16/plexpy/blob/login-page/plexpy/pmsconnect.py#L1056

drzoidberg33:
> was the current activity a track?

Yes.

JonnyWong16:
Ok, I think I got it. Background tasks seem to be working now.

JonnyWong16:
Ok, I'm going to call this done.

Hellowlol:
We should should consider caching the images so we dont have any additional http overhead to pms everytime.

drzoidberg33:
A potential issue is that shared users are able to view metadata for items that they aren't allowed to access. For instance, restricting movies for a user to family only ratings will still allow them to see all restricted items in PlexPy. Same issue with sharing labels.

Slightly tricky to get this working with history however direct calls to Plex will be fine as long as you're using the correct token.

JonnyWong16:
How would we cache? To disk? To the browser?

Can we make the assumption that if an item is in a user's history, then they had the appropriate permissions at the time they viewed it? If the particular library has been unshared and they viewed it previously it should still count towards their stats.

Hellowlol:
We could use the same way htpc manager does. Hash the URL and store in plexpys cache dir, check cache before requesting a new image from pms and change from a query URL to /image/1112/100/150 so we don't bust the cache. Isn't there a way check the shared libs while refreshing the users?

JonnyWong16:
Where are the images stored?

Would it really make much of a difference to store the shared library IDs in the database? It only hits the Plex API once per login to check the libraries.

Hellowlol:
The images are stored in pms atm so they are transcoded on the fly or served from pms cache.

JonnyWong16:
I see how you did it here: https://github.com/Hellowlol/HTPC-Manager/blob/master2/htpc/helpers.py#L54

Hash the image URL (`/library/metadata/28103/thumb/1460251799`) and store the `.jpg`. The problem is that the timestamp in the thumbnail URL changes frequently (on each server maintenance?). So hash without the timestamp I guess?

How often/when do you clear the cache for HTPC-Manager?

This won't be in this PR.

Hellowlol:
I never clear the cache, it can be cleared manually from the gui or by deleting the cache folder. The plex and kodi modules also have options to download every simple images it need to refresh the cache. (URL has to be written in manually, it only exist since I'm lazy and don't want the waiting time that scrolling does.) I can send a pr for it if you wish.

JonnyWong16:
Sure, this should be a very simple PR. Just the backend, I'll figure out the button to clear the cache in the gui.

JonnyWong16:
~~Bug. Guest can still login after removing them from Plex. Need to check user **and** server token on login.~~

Fixed in 3bf5a9e.

JonnyWong16:
@drzoidberg33 Implemented filter/mask metadata based on content rating and labels.

drzoidberg33:
~~Cool, seems to work well except how do you refresh the permissions for a user? For example I restricted a section to test, then unrestricted it and did a library refresh with the admin user but doesn't seem to work.~~

~~Oh and tried a user refresh too, but that didn't fix it either.~~

Ignore me, I was refreshing the wrong PlexPy tab :panda_face: :palm_tree:

JonnyWong16:
Should have rebased before merging....not after...Oops...

Arcanemagus:
Should this be working on `dev` now? I'm unable to log in with a test user, can file a new issue if it should be working.

JonnyWong16:
It sound be working on dev. Did you enable guest access for the user?

Arcanemagus:
"Allow Guest Access to PlexPy" is enabled in the Access Control section, and all users tested have "Allow Guest Access" enabled.

All that I see in the logs is:

```
Invalid login attempt from 'arcanemagus-test'.
```

Arcanemagus:
A [few](https://github.com/drzoidberg33/plexpy/commit/b6461f4f9ebc59d852df7e0871c562e7d0c4afb7) [fixed](https://github.com/drzoidberg33/plexpy/commit/a1821fabf9d2131a895ce80b51209a5a3911f56f) [bugs](https://github.com/drzoidberg33/plexpy/commit/3e6f5ac70e59ae765edb0a82cb1291936e6a0194) later and everything is working fine, thanks @JonnyWong16!

PR #689, opened 2016-05-05
cache images, download log, remove memory log

Hellowlol:


plexpy/webserve.py
```diff
@@ -1915,21 +1933,98 @@ def get_item_children(self, rating_key='', **kwargs):
...
+            logger.debug('No image input received')
+            return
 
-        if result:
-            cherrypy.response.headers['Content-type'] = result[1]
-            return result[0]
-        else:
-            return None
+        if ratingkey and not img:
+            img = '/library/metadata/%s/thumb/1337' % ratingkey
```
JonnyWong16, line 1950: Lol. I see what you did there.

plexpy/webserve.py
```diff
@@ -1915,21 +1933,98 @@ def get_item_children(self, rating_key='', **kwargs):
...
+
+        i = img.rsplit('/', 1)
+        i += [width, height]
+        img_string = '/'.join(i)
+        fp = hashlib.md5(img_string).hexdigest()
+        fp += '.jpg'  # we want to be able to preview the thumbs
```
JonnyWong16, line 1955: I've found that some of my cached images are `.png`. Double check? Maybe I'm wrong?

plexpy/webserve.py
```diff
@@ -1915,21 +1933,98 @@ def get_item_children(self, rating_key='', **kwargs):
...
+            # the image does not exist, download it from pms
+            try:
+                pms_connect = pmsconnect.PmsConnect()
+                result = pms_connect.get_image(img, width, height)
+                cherrypy.response.headers['Content-type'] = result[1]
+
+                if result and result[0]:
+                    if 'indexes' not in img:
+                        with open(ffp, 'wb') as f:
+                            f.write(result[0])
```
JonnyWong16, line 1977: Isn't it still possible that this will cache a smaller image? e.g. the small thumbnail from the popup on the history table, then serve that smaller image for the larger posters elsewhere?

plexpy/webserve.py
```diff
@@ -1230,24 +1247,25 @@ def get_notification_log(self, **kwargs):
...
-            return json.dumps({'message': 'no data received'})
+        return {'message': result}
 
     @cherrypy.expose
     @requireAuth(member_of("admin"))
     def clearLogs(self):
-        plexpy.LOG_LIST = []
-        logger.info(u"Web logs cleared")
+        try:
+            open(os.path.join(plexpy.CONFIG.LOG_DIR, 'plexpy.log'), 'w').close()
```
JonnyWong16, line 1270: I liked the old way where clearing the logs would only clear it from the web UI, and not actually clear the file.

There have been times when I've had to help someone and they said they cleared the logs (which only cleared them from the web UI), but thankfully the log file was still complete.

plexpy/pmsconnect.py
```diff
@@ -1896,51 +1896,24 @@ def get_image(self, img=None, width=None, height=None, fallback=None):
...
-        if not 'indexes' in image_uri:
-            image_path, content_type = helpers.cache_image(image_uri)
 
-            if image_path and content_type:
-                return [open(image_path, 'rb'), content_type]
-
-        try:
+        if img:
+            uri = '/photo/:/transcode?url=http://127.0.0.1:32400%s' % img
             if width.isdigit() and height.isdigit():
```
JonnyWong16, line 1902: I think the uri _must_ include a width and height. I get `400 Bad Request` if I don't include the dimensions.

plexpy/webserve.py
Hellowlol, line 1955: It doesnt really matter, i just wanted a extension so they are showed in windows.

plexpy/webserve.py
Hellowlol, line 1977: No, the hashing are with the width and height, so when it checks it should raise NotFound and cache the correct image.

plexpy/webserve.py
Hellowlol, line 1270: I see your point, but we cant protect the user from everything, there is already a warning when we try to delete the log. There isnt anyway to "clear" the weblog since there are no more weblog. It reads the real log. I can change this back if you want to, but i think its nice to not hold every logging in memory, even if the log rolls the memory log keep getting larger and larger

plexpy/pmsconnect.py
Hellowlol, line 1902: :( Ill do some digging.

plexpy/webserve.py
Hellowlol, line 1950: :P

plexpy/webserve.py
JonnyWong16, line 1977: Ah, I missed that on L1947. Is it better to cache one larger image and resize as necessary, instead of caching one image at every size?

plexpy/webserve.py
JonnyWong16, line 1955: But does serving a `.png` as `image/jpeg` work (L1960 below)?

plexpy/webserve.py
Hellowlol, line 1977: I dont see the point to cache the larger and resize on the fly unless you want to do it locally thus installing pil/pillow. I dont think we should do that.

plexpy/webserve.py
Hellowlol, line 1955: Hasnt barfed yet :P

plexpy/webserve.py
Hellowlol, line 1270: Maybe we can roll over the log ^^

JonnyWong16, line 1270: I guess if the user clears their logs, and they need help, then it's their fault.

It's not like everyone posts their logs when they have problems anyways...So really no difference =P

plexpy/webserve.py
```diff
@@ -1915,21 +1933,98 @@ def get_item_children(self, rating_key='', **kwargs):
...
-        if result:
-            cherrypy.response.headers['Content-type'] = result[1]
-            return result[0]
-        else:
-            return None
+        if ratingkey and not img:
+            img = '/library/metadata/%s/thumb/1337' % ratingkey
+
+        i = img.rsplit('/', 1)
+        i += [width, height]
```
JonnyWong16: Image url split but the timestamp is still there when you join it back together.

Hellowlol: Thanks!

plexpy/webserve.py
```diff
@@ -1180,13 +1183,28 @@ def getLog(self, start=0, length=100, **kwargs):
             search_value = kwargs.get('search[value]', "")
 
         if 'search[regex]' in kwargs:
-            search_regex = kwargs.get('search[regex]', "")
+            search_regex = kwargs.get('search[regex]', "") # should this be removed?
+
+        filt = []
+        fa = filt.append
+        with open(os.path.join(plexpy.CONFIG.LOG_DIR, 'plexpy.log')) as f:
```
JonnyWong16, line 1196: I feel like this entire section should be moved to `log_reader.py` to make things cleaner.

I'm just being picky now.

plexpy/webserve.py
```diff
@@ -1180,13 +1183,28 @@ def getLog(self, start=0, length=100, **kwargs):
...
+            for l in reversed(f.readlines()):
+                try:
+                    temp_loglevel_and_time = l.split('- ')
```
JonnyWong16, line 1199: I think this leaves a trailing space on the log time. `split(' - ')` instead?

Again being picky.

JonnyWong16:
I haven't actually tested anything. I've just been commenting by running the code in my head.

plexpy/webserve.py
```diff
@@ -1915,21 +1933,98 @@ def get_item_children(self, rating_key='', **kwargs):
 
     @cherrypy.expose
     @requireAuth()
-    def pms_image_proxy(self, img='', width='0', height='0', fallback=None, **kwargs):
+    def pms_image_proxy(self, img='', ratingkey=None, width='0', height='0', fallback=None, **kwargs):
+        """ Grabs the images from pms and saved them to disk """
 
-        pms_connect = pmsconnect.PmsConnect()
-        result = pms_connect.get_image(img, width, height, fallback)
+        if not img or ratingkey:
```
JonnyWong16: `not img or not ratingKey`

plexpy/webserve.py
```diff
@@ -1915,21 +1933,98 @@ def get_item_children(self, rating_key='', **kwargs):
...
+        fp = hashlib.md5(img_string).hexdigest()
+        fp += '.jpg'  # we want to be able to preview the thumbs
+        c_dir = os.path.join(plexpy.CONFIG.CACHE_DIR, 'images')
+        ffp = os.path.join(c_dir, fp)
+
+        if not os.path.exists(c_dir):
+            os.mkdir(c_dir)
+
+        try:
+            if 'indexes' in img:
```
JonnyWong16, line 1963: Very slight optimization if we check `if 'indexes' in img` earlier? Probably negligible.

Hellowlol, line 1963: We can't unless we refactor the image download. That's why I raise the exception manually because I know it gonna fail anyway.

JonnyWong16:
Can you rebase?

PR #713, opened 2016-05-18
fix https://github.com/drzoidberg33/plexpy/issues/705#issuecomment-21…

Hellowlol:
…9927893

Can you test and verify?


JonnyWong16:
Yep, looks right now. Thanks.

PR #729, opened 2016-05-24
Add log level filter for plex server and media logs

sanderploegsma:
Since the plex logs are overflowing with debug logs I liked being able to filter based on log level.


sanderploegsma:
On second thought, this doesn't add much functionality, as the logs contain only the last 1000 lines, just enough for one minute of activity when Plex is actively streaming.

JonnyWong16:
~~This filters everything at the level and above?~~

Figured it out. Works nicely. Thanks.

PR #735, opened 2016-05-26
Reflect path suggested in installation guide

nortron:
The installation guide suggests the install path as "/usr/local/share" for FreeBSD. This is also stated in the daemon guide but the daemon file to be copied does not reflect that suggested path. This fixes that descrepency (in tandem with the changes I just made to the wiki page).


JonnyWong16:
Thanks.

PR #741, opened 2016-06-02
allow refresh of images

Hellowlol:
Allow refresh of images and fix bug where a disabled cache still would
use the cache

### DIFF
--- a/plexpy/webserve.py
+++ b/plexpy/webserve.py
@@ -749,16 +749,16 @@ class WebInterface(object):
 
             Returns:
                 json:
-                    {"child_count": null, 
-                     "count": 887, 
-                     "do_notify": 1, 
-                     "do_notify_created": 1, 
-                     "keep_history": 1, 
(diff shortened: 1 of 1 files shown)


JonnyWong16:
That was fast...

Hellowlol:
Well take a look at the diff, wasnt any work really :P. Dunno if it should be a button somewhere or if this something they have to do manually.

JonnyWong16:
There should probably be a button. That's going to be quite a task to add everywhere.

Hellowlol:
Then don't. I wouldn't be arse with that. Just add a button in the view show/movie and/or write a general function that recaches every image of every show/ep/movie

JonnyWong16:
That's probably even more difficult when the files are hashed.

Hellowlol:
Ah true, i just thought we compare the hash but i forgot that we also include the filesize etc. That would be major pain :P

PR #773, opened 2016-07-07
Added Hipchat integration notifier.

aboron:
Uses chat room BYO integration, copy URL from the box on Hipchat that says "Send messages to this room by posting to this URL" (Link provided in settings for example)


plexpy/notifiers.py
```diff
@@ -2721,3 +2748,71 @@ def return_config_options(self):
...
+
+    def __init__(self):
+        self.apiurl = plexpy.CONFIG.HIPCHAT_URL
+        self.color = plexpy.CONFIG.HIPCHAT_COLOR
+
+    def notify(self, message, subject):
+        if not message or not subject:
+            return
+        
+        text = '(plex) ' + subject.encode('utf-8') + ': ' + message.encode('utf-8')
```
JonnyWong16: Any of the notifications agents where subject and message are sent in one string should have the option to turn off the subject line. See Telegram/Slack/Twitter/Facebook/etc.

Also, why is `(plex)` added to the beginning?

aboron: Added the subject option.  Thanks for catching the (plex) emoticon tag, I threw that in early and forgot I hadn't even made it optional without any explanations.  I added an option to control that also.

Hipchat lets you define custom emoticons and the messages look nice with a plex emoticon defined and used at the start of the lines like this:
message_example

JonnyWong16: I see. Is `(plex)` common across Hipchat? Or do you create your own emoticon? What if someone creates one but doesn't name it `(plex)`? It might be better to have it as a string instead of a integer `0/1`, so I could use `(plexpy)` instead or something.

```
'HIPCHAT_EMOTICON': (str, 'Hipchat', ''),

...

self.emoticon = plexpy.CONFIG.HIPCHAT_EMOTICON

...

if self.emoticon:
    text = self.emoticon + ' ' + text

...

{'label': 'Hipchat Emoticon',
 'value': self.emoticon,
 'name': 'hipchat_emoticon',
 'description': 'Include an emoticon tag (e.g. (plexpy)) at the beginning of all notifications. Leave blank for none. '
                'Create a custom emoticon <a href="' + helpers.anon_url('https://www.hipchat.com/emoticons/') + '" target="_blank">here</a>.',
 'input_type': 'text'
 }
```

aboron: Good point, and done.

JonnyWong16:
Cool. I'll test this later.

plexpy/notifiers.py
```diff
@@ -2721,3 +2748,94 @@ def return_config_options(self):
...
+                          'description': 'Your Hipchat integration URL. You can get a key from'
+                                         ' <a href="' + helpers.anon_url('https://www.hipchat.com/addons/') + '" target="_blank">here</a>.',
+                          'input_type': 'text'
+                          },
+                         {'label': 'Hipchat Color',
+                          'value': self.color,
+                          'name': 'hipchat_color',
+                          'description': 'Background color for the message.',
+                          'input_type': 'select',
+                          'select_options': {'yellow': 'yellow', 'green': 'green', 'red': 'red', 'purple': 'purple', 'gray': 'gray', 'random': 'random'}
```
JonnyWong16, line 2825: You need a blank/no colour option.

plexpy/notifiers.py
```diff
@@ -2721,3 +2748,94 @@ def return_config_options(self):
...
+        
+        if self.incl_subject:
+            text = subject.encode('utf-8') + ': ' + message.encode('utf-8')
+        else:
+            text = message.encode('utf-8')
+
+        if self.emoticon:
+            text = self.emoticon + ' ' + text
+
+        data = {'color': self.color,
```
JonnyWong16: Will this still send if color is blank?

plexpy/notifiers.py
aboron, line 2825: The api will use 'gray' if it is not specified.  Would it be more appropriate if I set that as the default in config.py?

JonnyWong16, line 2825: Is this text colour? or background colour?

I would go with a blank default.

plexpy/notifiers.py
aboron: If I send a json POST with a blank color value it gives an error, but if the color option is not sent at all, it works fine and uses grey.  Examples of output colors:
background

JonnyWong16: That's what I though. Set the default blank then add a condition:

```
if self.color:
    data['color'] = self.color
```

JonnyWong16:
Works. Thanks!

PR #779, opened 2016-07-13
Adds posters option for Hipchat and metadata formatting functions.

aboron:
Example poster notification ( Closed and Open )
closed
open


plexpy/notifiers.py
```diff
@@ -609,12 +609,78 @@ def send_notification(agent_id, subject, body, notify_action, **kwargs):
...
+            self.caption = 'View on The Movie Database'
+        elif self.metadata.get('imdb_url',''):
+            self.caption = 'View on IMDB'
+        elif self.metadata.get('lastfm_url',''):
+            self.caption = 'View on Last.fm'
+        return self.caption
+
+    def get_title(self, divider = ''):
+        self.title = None
+        if self.metadata['media_type'] == 'movie':
```
JonnyWong16: You can use `self.media_type` or get rid of `self.media_type` above.

plexpy/notifiers.py
```diff
@@ -609,12 +609,78 @@ def send_notification(agent_id, subject, body, notify_action, **kwargs):
...
+            self.caption = 'View on TheTVDB'
+        elif self.metadata.get('themoviedb_url',''):
+            self.caption = 'View on The Movie Database'
+        elif self.metadata.get('imdb_url',''):
+            self.caption = 'View on IMDB'
+        elif self.metadata.get('lastfm_url',''):
+            self.caption = 'View on Last.fm'
+        return self.caption
+
+    def get_title(self, divider = ''):
```
JonnyWong16: Default `divider = '-'`?

I don't know what looks better.

```
S1 - E1
S1-E1
S1E1
S1 E1
```

Facebook is the only agent that uses

```
S1 · E1
```

plexpy/notifiers.py
```diff
@@ -2754,32 +2723,82 @@ class HIPCHAT(object):
...
+            poster_link = pretty_metadata.get_poster_link()
+            caption = pretty_metadata.get_caption()
+            title = pretty_metadata.get_title()
+            subtitle = pretty_metadata.get_subtitle()
+            plex_url = pretty_metadata.get_plex_url()
+
+            card = {'title': title,
+                    'format': 'medium',
+                    'style': 'application',
+                    'id': '41c354f0-69a5-4251-a61c-c6a8daf8e2f6'}
```
aboron: I used random GUID generation.  It is a required field, but seems to have no effect that I can discern.  The api docs say:

> An id that will help HipChat recognise the same card when it is sent multiple times
> Valid length range: 1 - unlimited.

plexpy/notifiers.py
aboron: Yep, diff problem on my end. I'm testing on a different box than I'm committing from. :-(
I am running with `divider = '-'`on the testing box.  I have no preferences on the shorthand.

plexpy/notifiers.py
JonnyWong16: Then we should probably generate a random one.

```
import uuid

...

    'id': uuid.uuid4().hex
```

plexpy/notifiers.py
```diff
@@ -2754,32 +2724,82 @@ class HIPCHAT(object):
     def __init__(self):
         self.apiurl = plexpy.CONFIG.HIPCHAT_URL
         self.color = plexpy.CONFIG.HIPCHAT_COLOR
-        self.incl_subject = plexpy.CONFIG.HIPCHAT_INCL_SUBJECT
         self.emoticon = plexpy.CONFIG.HIPCHAT_EMOTICON
+        self.from_subject = plexpy.CONFIG.HIPCHAT_FROM_SUBJECT
+        self.incl_pmslink = plexpy.CONFIG.FACEBOOK_INCL_PMSLINK
+        self.incl_poster = plexpy.CONFIG.FACEBOOK_INCL_POSTER
```
JonnyWong16: `FACEBOOK` typo these two lines.

plexpy/notifiers.py
```diff
@@ -2754,32 +2724,82 @@ class HIPCHAT(object):
...
+
+        text = message.encode('utf-8')
+
         if self.incl_subject:
-            text = subject.encode('utf-8') + ': ' + message.encode('utf-8')
-        else:
-            text = message.encode('utf-8')
+            text = subject.encode('utf-8') + ': ' + text
 
         if self.emoticon:
```
JonnyWong16: Can't have emoticon if "Include Poster" is checked.

The notification still appends `(plexpy)` to the beginning.

plexpy/notifiers.py
```diff
@@ -2754,32 +2724,82 @@ class HIPCHAT(object):
...
             text = self.emoticon + ' ' + text
 
-        data = {'message': text,
-                'notify': 'false',
-                'message_format': 'text'}
-
         if self.color:
             data['color'] = self.color
 
+        if self.from_subject:
```
JonnyWong16: If `HIPCHAT_INCL_SUBJECT` and `HIPCHAT_FROM_SUBJECT` are both checked, the subject line shows up twice. I might be better to only have one setting.

If `HIPCHAT_INCL_SUBJECT` is enabled then have the subject as the "From" line?

aboron: I prefer it on the from line in either format, so I'll consolidate that. Plus the other items from above.

JonnyWong16:
Alright, looks good. Thanks.

PR #838, opened 2016-09-08
Attempt to verify the PID in an existing PID file before giving up.

spolyack:
If the PID doesn't map to a running process, then we can simply ignore the presence of the PID file and overwrite it with the current (new) PID later.

This allows plexpy to start correctly after a system crash or power loss.


JonnyWong16:
Thanks, but please submit pull requests to the `dev` branch.

PR #839, opened 2016-09-12
Quick Fix for iPhone 6/7 Plus on PlexPy Website

XusBadia:
I changed the max-device-width from 400px to 450px so it adjusts to the iPhone 6/7 Plus screen and doesn't leave a gap on the right of the screen.


JonnyWong16:
Thanks, but please submit pull requests to the `dev` branch.

PR #842, opened 2016-09-18
Quick Fix for iPhone 6/7 Plus

XusBadia:
I changed the max-device-width from 400px to 450px so it adjusts to the iPhone 6/7 Plus screen and doesn't leave a gap on the right of the screen.

This should now be in the correct branch (sorry about that!)


JonnyWong16:
Thanks.

PR #856, opened 2016-10-02
fix for image proxy for api.

Hellowlol:


JonnyWong16:
I see, you proxied the proxy. Thanks!

PR #870, opened 2016-10-13
Incremented table width to 100%

XusBadia:
Table width (header and body/back) incremented to 100% but adding max-width so it isn't too wide in wide-screen monitors.


JonnyWong16:
Thanks.

PR #898, opened 2016-11-03
fix result type, fixup responses from the ui to api

Hellowlol:
todo unify json responses from webserve.py

### DIFF
--- a/plexpy/api2.py
+++ b/plexpy/api2.py
@@ -47,7 +47,7 @@ class API2:
         self._api_cmd = None
         self._api_apikey = None
         self._api_callback = None  # JSONP
-        self._api_result_type = 'failed'
+        self._api_result_type = 'error'
         self._api_profileme = None  # For profiling the api call
         self._api_kwargs = None  # Cleaned kwargs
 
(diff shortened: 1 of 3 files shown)


plexpy/api2.py
```diff
@@ -524,13 +506,13 @@ def _api_run(self, *args, **kwargs):
...
-            if self._api_debug:
+            try:
                 result = call(**self._api_kwargs)
-            else:
-                try:
-                    result = call(**self._api_kwargs)
-                except Exception as e:
-                    logger.error(u'PlexPy APIv2 :: Failed to run %s %s %s' % (self._api_cmd, self._api_kwargs, e))
+            except Exception as e:
+                if self._api_debug:  # check this with j
```
Hellowlol: @JonnyWong16 Whats your take this on? I would like to get the traceback in the browser if the debug param is used with the api.

JonnyWong16: Fine with me, only if `debug=1`.

plexpy/webserve.py
```diff
@@ -329,6 +329,7 @@ def get_recently_added(self, count='0', **kwargs):
     @cherrypy.expose
     @cherrypy.tools.json_out()
     @requireAuth(member_of("admin"))
+    @addtoapi()
     def delete_temp_sessions(self, **kwargs):
```
JonnyWong16, line 333: Needs a docstring.

```
Flush out all of the temporary sessions in the database.
```

And run `docs_md` to  update the API docs.

Hellowlol:
Should be good now.

PR #909, opened 2016-11-27
Allow certificates to be chained certificates

simonarons:
There are issues with SSL cert error on Android when using certs from certain issuers (e.g. letsencrypt). Allowing the cert to be a chained cert solves this. Doesn't affect the use of SSL certs in any other way.


JonnyWong16:
Submit all pull requests to the dev branch.

Also please update the help text in the settings.
https://github.com/JonnyWong16/plexpy/blob/dev/data/interfaces/default/settings.html#L463

```
The location of the SSL certificate or certificate chain.
```

PR #913, opened 2016-12-03
Add requirements.txt

Cabalist:
This small pull requests adds a requirements.txt file to the repository.  Currently the python dependencies are distributed alongside the plexpy code in the lib/ directory.  I believe this is for convenience of installation.  However this presents a number of issues including maintenance and licensing issues.  

Everything in the lib/ dir is included in this requirements.txt except for the following files:

```
certgen.py
hashing_passwords.py
```

`certgen.py` appears to be from the pyOpenSSL Examples folder and is being distributed without it's required License
`hashing_passwords.py` is from the blog post here:  http://exyr.org/2011/hashing-passwords/

I believe both these files would be better adopted into the plexpy code repo instead of the lib dir which contains 3rd party code.

This file can be adapted into a setup.py file down the road and make distribution and installation of this project as easy as `pip install plexpy`

Let me know your opinion on this and if you are interested in getting this package on PyPI.


JonnyWong16:
Yes, please incorporate those two files into the PlexPy code if you can. 

I changed one line in the CherryPy library to work with PlexPy, I can't remember if I modified any of the other libraries. I don't think I did. 

Libraries are included so that you can just download one zip file and run PlexPy right away. Not everyone wants to, or know how to, work with `pip`. Most people know how to download a file and double click.  It would be nice if both options could be maintained.

Cabalist:
I'll take a look at the libraries to see if/where things have changed.  I think it would be interesting to get this all setup so that it can be a more standard install on Windows and Mac as well as Linux.

Here is a (sadly incomplete) list of some options that can be accomplished with a properly configured setup.py.  http://docs.python-guide.org/en/latest/shipping/freezing/

Providing a 'frozen' app would allow you to put a download link in the readme and make sure the environment that PlexPy is running in is clean and doesn't have any conflicts.  The end-user wouldn't need pip or even Python for the app to run.  

Anyways! I'll also incorporate those two files in plexpy dir and update this pull request.

JonnyWong16:
I like the idea of a 'frozen' app.

Cabalist:
I would say Py2exe for windows, py2app for OSX and `pip` for the Linux & BSD folks.  
I can work on the last two but windows is a challenge to me and I'll probably leave that to someone else.  😉

Cabalist:
This works for me but could really benefit from some more verification.

PR #914, opened 2016-12-03
Pep8 cleanups, small (really small) bug fixes, cleanup imports

Cabalist:
Hello!

I am reading though your code as an exercise and made some small edits along the way.  Almost all of this is PEP8 related and generated by my linter.  No behavior should have changed at all.  The biggest changes I made were:

1. Imports have been organized and trimmed down
2. Catchall exceptions are now being logged. 
3. Mixed tabs and spaces have been converted to spaces
4. Unnecessary escape characters were removed.
5. Unreachable `return` statements removed
6. Unreachable `except` clauses removed when they are redundant as well
7. Whitespace

After reviewing the code I think it would be very possible to get this working in Python3 with little trouble.  Any interest in some Pull Requests to that end?

### DIFF
--- a/plexpy/plexivity_import.py
+++ b/plexpy/plexivity_import.py
@@ -1,4 +1,5 @@
-﻿# This file is part of PlexPy.
+﻿# coding=utf-8
+# This file is part of PlexPy.
 #
 #  PlexPy is free software: you can redistribute it and/or modify
 #  it under the terms of the GNU General Public License as published by
@@ -13,12 +14,11 @@
 #  You should have received a copy of the GNU General Public License
 #  along with PlexPy.  If not, see <http://www.gnu.org/licenses/>.
 
-import arrow
 import sqlite3
 from xml.dom import minidom
 
-import plexpy
-import activity_pinger
+import arrow
+
 import activity_processor
 import database
 import helpers
@@ -32,8 +32,9 @@ def extract_plexivity_xml(xml=None):
     clean_xml = helpers.latinToAscii(xml)
     try:
         xml_parse = minidom.parseString(clean_xml)
-    except:
+    except Exception as e:
         logger.warn(u"PlexPy Importer :: Error parsing XML for Plexivity database.")
+        logger.exception(e)
         return None
 
     # I think Plexivity only tracked videos and not music?
@@ -234,6 +235,7 @@ def extract_plexivity_xml(xml=None):
 
     return output
 
+
 def validate_database(database=None, table_name=None):
     try:
         connection = sqlite3.connect(database, timeout=20)
@@ -243,8 +245,9 @@ def validate_database(database=None, table_name=None):
     except ValueError:
         logger.error(u"PlexPy Importer :: Invalid database specified.")
         return 'Invalid database specified.'
-    except:
+    except Exception as e:
         logger.error(u"PlexPy Importer :: Uncaught exception.")
+        logger.exception(e)
         return 'Uncaught exception.'
 
     try:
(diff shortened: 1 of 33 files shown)


Hellowlol:
I like it.

JonnyWong16:
Thanks! I just did a quick skim through and it all looks good, but I don't like how it added `logger.exception(e)` everywhere. Not everything needs to be sent to the logger.

As for Python 3 support, you're welcome to add it. In just worried about the dependencies as I don't think all the libraries are available on python 3. I've been meaning to do it but I'd rather work on other features.

Cabalist:
Yeah.  I went back and forth on the logging.  Would those be better off as Debug logs?  I hate just throwing away exceptions.  

Cool.  I'll start chipping away at Py3 support.

Quick questions:
What is the minimum Python2 version you would like to support?   (recommend 2.7)
What is the minimum Python3 version you would like to support?  (recommend 3.5)
Do you have a test suite?  I see some references to travis-ci in `Contributing.md` but no other info about it.  
What platforms are currently supported?

plexpy/plexivity_import.py
```diff
@@ -32,8 +32,9 @@ def extract_plexivity_xml(xml=None):
     clean_xml = helpers.latinToAscii(xml)
     try:
         xml_parse = minidom.parseString(clean_xml)
-    except:
+    except Exception as e:
         logger.warn(u"PlexPy Importer :: Error parsing XML for Plexivity database.")
+        logger.exception(e)
```
JonnyWong16, line 37: For example these can be changed to:
```
logger.error(u"PlexPy Importer :: Error parsing XML for Plexivity database: %s" % e)
```

Instead of having two logger messages.

Hellowlol:
I think the exception should describe what it tried to do. Fx right now we have a exception that only log a unhandled exception. Kinda useless if you ask me, I like that log.exception atleast provides a trace back so I can see where it went wrong but something like tried to execute sql ... but failed and the trackball on the next line.

Hellowlol:
Also there is not tests. So before porting the code we should have some tests. I suggest pytest to reduce boilerplate

JonnyWong16:
Python 2.7/3.5

We don't have tests. Also something I've been meaning to do, but too lazy to do it. I want to get some tests and CI going.

Not sure what you mean by platforms. All?

Cabalist:
Platforms = Windows, Mac, Linux, etc.  

This can be a challenge sometimes depending on how the code is accessing the OS but now that you mention it I haven't seen any workarounds in the code so there probably isn't too much that is platform dependent.  

@Hellowlol 👍 Pytest

JonnyWong16:
Then all platforms need to be supported (this includes a bunch of NAS devices).

There are a few Windows workarounds. Off the top of my head in `plexpy.shutdown(restart=True)` and in `plexpy.notifiers.SCRIPTS.notify` (I think).

Hellowlol:
The workaround of scripts is fine we just need something handle py3. It's a minor change.

Cabalist:
The NASs I am aware of all run Linux so we should be good there.  

As for the exceptions there is a lot of silent failures in the code that should probably be more explicit.  Logging is a good step but it is totally up to you on how to do it.  

I'd like the Expected exceptions to be caught and gracefully handled and the Uncaught/Unexpected exceptions to make more noise so that they can be fixed early.

Hellowlol:
Also while we are speaking about improvements it think we should make thing easier to mock, accepting a data value.
Fx: https://github.com/JonnyWong16/plexpy/blob/master/plexpy/api2.py#L122
```
logfile = data or open(os.path.join(plexpy.CONFIG.LOG_DIR, logger.FILENAME)).readlines()
```
Json responses in webserve.py shoul be unified. Atm the api guesses and tries to get it right, im sure it fails on some calls :P

PR #915, opened 2016-12-04
Python3 Support

Cabalist:
Sooo.....  I don't expect this to get merged as it doesn't currently support Python3 but I am calling attention to it so that others can build on it.   

This pull request takes care of the low hanging fruit when it comes to Python3 support.  This should be almost everything necessary.  The `six` library is integrated and updates to all the dependencies for Python3 compatibility is included in the `requirements.txt`.  The blocking issue that arose is that PlexPy in it's current state has a great deal of circular dependencies.  

For example:

logger.py:
```python
import plexpy
import helpers
```

helpers.py:
```python
import plexpy
import logger
from api2 import API2
```

This 'works' in Python2 because it is using ambiguous relative imports.  In Python3 these were removed and explicit relative imports ( ex.  `from foo import bar` or `from . import bar`) are the only ones allowed.

I do not currently know enough about the structure of this project to undertake a major refactor like this.  However someone with more familiarity should be able to get it done without too much issue.  I have faith!  This last hurdle is all you need for Python3 support.

Meanwhile this pull request should function entirely the same as the existing version.  It runs in Python 2.7 without issue.  Also by coercing the imports you can get it to run in Python3 without observable issue.  But I couldn't get it to run in both without fixing the structure.


PlexPy.py
```diff
@@ -21,20 +16,20 @@
 #  You should have received a copy of the GNU General Public License
 #  along with PlexPy.  If not, see <http://www.gnu.org/licenses/>.
 
-import os
-import sys
-
-# Ensure lib added to path, before any other imports
-sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'lib/'))
```
Cabalist, line 28: This is the only breaking change and I removed it for development.  Instead of looking in the provided `lib/` dir this looks at the installed libraries.  When this Pull request is up to snuff I imagined updating the `lib/` from the installed library and not dirtying an already nit-picky commit.

JonnyWong16:
Message me on [Discord](https://discord.gg/36ggawe).

PR #951, opened 2017-01-23
Small example of making a request using curl

hamiltont:
Not sure if this is useful enough to be included, but it would have saved me about 2min since i thought the API was POST-only

### DIFF
--- a/API.md
+++ b/API.md
@@ -30,6 +30,24 @@ General optional parameters:
     debug:      1
 ```
 
+### Request Example
+
+Making a request to PlayPy using the [curl](https://curl.haxx.se/) tool is straightforward
+
+```
+your_name$ curl "http://<hostname>:<port>/api/v2?apikey=<apikey>&cmd=get_home_stats"  \
+	| python -m json.tool        `# Make the output pretty`
+{
+    "response": {
+        "data": [
+            {
+                "rows": [],
+                "stat_id": "top_movies",
+                "stat_type": "total_plays"
+            }
+            ...<snip>...
+```
+
 ## API methods
 
 ### arnold


hamiltont:
ps - targeted dev branch. Is that correct?

API.md
```diff
@@ -30,6 +30,24 @@ General optional parameters:
     debug:      1
 ```
 
+### Request Example
+
+Making a request to PlayPy using the [curl](https://curl.haxx.se/) tool is straightforward
```
Arcanemagus, line 35: * `PlayPy` -> `PlexPy`
* Put a `:` at the end of the line

API.md
```diff
@@ -30,6 +30,24 @@ General optional parameters:
...
+
+```
```
Arcanemagus, line 37: If you set the language to `ShellSession` it will do some highlighting for you:

\`\`\`ShellSession

API.md
```diff
@@ -30,6 +30,24 @@ General optional parameters:
...
+your_name$ curl "http://<hostname>:<port>/api/v2?apikey=<apikey>&cmd=get_home_stats"  \
```
Arcanemagus, line 38: The `your_name` is implied from the `$`, no need to include it.

Arcanemagus (changes requested):

Hellowlol:
There shouldn't be any direct pulls to the API md. The page is made by the API. See https://github.com/JonnyWong16/plexpy/blob/master/plexpy/api2.py#L344

Adding debug=1 will pretty print and sort the keys for you.
When I think about it, it would be cool to have a API builder like sickbeard did.

hamiltont:
oop, good call @Hellowlol . I'll just close this PR and may open a new one later on that file

PR #963, opened 2017-02-09
Add Discord notification agent

Svampen:
As Discord has opened up for [webhooks](https://support.discordapp.com/hc/en-us/articles/228383668-Intro-to-Webhooks) I thought it a good idea to update this great tool with the ability to post Plex notifications to a Discord channel.

The code is basically copied from the Slacks implementation with minor changes to adjust for layout of the message.
I also allowed for  synopses (poster info) to be present even if user don't want the poster image (as it requires a imgur account and without the original poster the message looks horrible)


JonnyWong16:
v2 already has Discord.

PR #974, opened 2017-02-21
Fix 404s for icon files and invalid paths

demonbane:
There are a number of files that throw 404s due to case-mismatch or missing paths. This fixes all of the ones I've been able to find. Renaming `Icon-*` to `icon-*` as well as removing `screen` from various paths that don't actually include that subdirectory.

Most notably, this should fix missing favorite icons for users on some iOS devices.


JonnyWong16:
Pull requests to the dev branch please.

PR #975, opened 2017-02-21
Fix 404s for icon files and invalid paths

demonbane:
There are a number of files that throw 404s due to case-mismatch or missing paths. This fixes all of the ones I've been able to find. Renaming Icon-* to icon-* as well as removing screen from various paths that don't actually include that subdirectory.

Most notably, this should fix missing favorite icons for users on some iOS devices.

The single deletion is for `Default-portrait@2x~iphone5.jpg` which is not referenced anywhere in the source and appears to be an old image.


JonnyWong16:
Thanks for this. I'll take a look when I have time.

PR #976, opened 2017-02-24
session_id notification type

hjone72:
Added session_id as notification option. This allows for scripts to kill a stream using the method used here:
http://forums.plex.tv/discussion/259921/kill-a-stream-button


JonnyWong16:
Thanks, I will be adding this to v2. I don't want to make any database changes to v1.

PR #982, opened 2017-02-27
Start day of the week changeable

Pbaboe:
Option in general settings to change the start day of the week to Monday (instead of the default of Sunday). This is only relevant for the Graph : Play by day of week.

Feathub issue : http://feathub.com/JonnyWong16/plexpy/+15

days-1

days-2


JonnyWong16:
Thanks for this. I'll take a look when I have time.

PR #984, opened 2017-02-28
{plex_url}

SRTwist:
Hallo, die {plex_url} VARIABLE scheint falsch zu sein.
Sorry ;-) i think there is an error in the URL of {plex_url} Variable, the missing piece is "?get="


Hellowlol:
You might want to write English so people understand you.

Issue #943, opened 2017-01-07
Bottom text not wrapping on history page


OmgImAlexis:
![screen shot 2017-01-08 at 2 52 54 am](https://cloud.githubusercontent.com/assets/6525926/21743230/ddc98446-d54d-11e6-801c-344b3d37ac07.png)

**What you did?**
Opened history page.
**What happened?**
Bottom text didn't wrap.
**What you expected?**
Bottom text to wrap.
**How can we reproduce your issue?**
Open history page on screen with width around 1440px.

JonnyWong16:
Fixed in v1.4.18.

PR #1006, opened 2017-03-26
Wrap history footer text

stephenp:
Fixes issue #943 

Apologies if I didn't do this right - this is my first PR for an open source project!

The CSS class "dataTables_info" has `white-space: nowrap;`, which is causing the issue.  I'm not sure where else this class is used, so I targeted the `#history_table_info` id instead.

The bootstrap grid classes are unaffected, so extra data will still be hidden as the viewport narrows.

Trailing spaces were also removed to clean this file up!

## Before 
before
## After
after

### DIFF
--- a/data/interfaces/default/css/plexpy.css
+++ b/data/interfaces/default/css/plexpy.css
@@ -404,6 +404,9 @@ textarea.form-control:focus {
     background-color: #fff;
     transition: background-color .3s;
 }
+#history_table_info {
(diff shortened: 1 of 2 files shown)


JonnyWong16:
Please make your changes on the `dev` branch and submit your pull requests to the `dev` branch.

Also, you might as well target all datatable footers with the class `.dataTables_info`.

PR #1007, opened 2017-03-26
Remove nowrap for footer text on history page

stephenp:
Fixes issue #943 

Making changes to `.dataTables_info` as [requested](https://github.com/JonnyWong16/plexpy/pull/1006#issuecomment-289323464).

The bootstrap grid classes are unaffected, so extra data will still be hidden as the viewport narrows.

## Before 
before
## After
after


JonnyWong16:
Fixed in fbe136a3507607cc2b92fb82d4a23188276178e8

PR #1021, opened 2017-04-22
Fixed math used to calculate human_duration

senepa:
There are 86400 seconds in a day


JonnyWong16:
My math is bad. Thanks!

PR #1025, opened 2017-04-27
Add support for looking up IPv6 addresses

dotsam:
Small changes to get_ip_address_details endpoint and isPrivateIP JS function to validate an IPv6 address correctly. geoip2 and ipwhois already support IPv6 addresses as-is.

### DIFF
--- a/data/interfaces/default/js/script.js
+++ b/data/interfaces/default/js/script.js
@@ -272,8 +272,15 @@ function isPrivateIP(ip_address) {
             return true;
         }
         return false;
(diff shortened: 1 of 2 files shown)


JonnyWong16:
I have already implemented this for v2 ca472ff5979d8b8e95261c4dcd66a24ddb4de28e

I'm missing  `fe80` and `fc00` though.

seanvree:
Could this by why I'm receiving this error:

"WebUI :: /history : Uncaught TypeError: isPrivateIP(...).then is not a function. (history_table.js:97)"

PR #1048, opened 2017-06-03
lossless compression of images saved 266KB

Fish2:
Save Bandwidth and speed up website load time but lossless completion of all images.

doing this saved 266KB docent sound like much but why not


JonnyWong16:
Some of those images are resources straight from Plex. I don't want to mess with those.

PR #1051, opened 2017-06-08
Update webstart.py

Napsterbater:
Allows Full Chain Certs.

See here https://github.com/JonnyWong16/plexpy/issues/660#issuecomment-249889711


JonnyWong16:
I have already added the chain certificate option to v2.

PR #1076, opened 2017-07-22
Added support for windows service

ghost:
Added support for running as a windows service so it does not fork the process on reboot.

### DIFF
--- a/plexpy/__init__.py
+++ b/plexpy/__init__.py
@@ -54,6 +54,7 @@ VERBOSE = True
 DAEMON = False
 CREATEPID = False
 PIDFILE = None
+WINDOWSSERVICE = False
 
 SCHED = BackgroundScheduler()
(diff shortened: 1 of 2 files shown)


plexpy/__init__.py
```diff
@@ -989,15 +990,17 @@ def shutdown(restart=False, update=False):
         args += ARGS
         if '--nolaunch' not in args:
             args += ['--nolaunch']
-        logger.info('Restarting PlexPy with %s', args)
 
         # os.execv fails with spaced names on Windows
         # https://bugs.python.org/issue19066
-        if os.name == 'nt':
+        if os.name == 'nt' and plexpy.WINDOWSSERVICE:
```
JonnyWong16, line 996:
```
if os.name == 'nt' and WINDOWSSERVICE:
```

PR #1139, opened 2017-11-08
Redirect to "home" after shutdown/restart

vertigo235:
In v2, "index" does not load anything

### DIFF
--- a/data/interfaces/default/shutdown.html
+++ b/data/interfaces/default/shutdown.html
@@ -51,7 +51,7 @@
 
     // Redirect to home page after countdown.
     function reloadPage() {
-        window.location.href = "${new_http_root}index";
+        window.location.href = "${new_http_root}home";
     }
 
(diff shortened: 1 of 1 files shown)


JonnyWong16:
No, index is correct.

vertigo235:
Interesting, I'm using a docker image so maybe that has something to do with it.

Hellowlol:
index is cherrypy special handler that defaults to the default page.

vertigo235:
Interesting, for v1, calling up /index, redirects to /home, but for me this doesn't happen with v2

vertigo235:
So I guess maybe the issue is somewhere else, I know this is a minor issue.

vertigo235:
I guess it's somewhere in here, I'll play with it later
https://github.com/JonnyWong16/plexpy/blob/v2/plexpy/webserve.py

PR #1151, opened 2017-11-20
Adds execute permission to fedora.centos and systemd init-scripts

wilmardo:
These two init-scripts were missing the execute permission, fixed it with this PR. 
If someone creates a symlinks from /etc/init.d/plexpy to the file in the git directory it won't work without the execute permission. 
The other init-scripts in the directory had the execute permission already set correctly.


JonnyWong16:
Please submit pull requests to the dev branch.

https://github.com/JonnyWong16/plexpy/blob/master/CONTRIBUTING.md#pull-requests

PR #1153, opened 2017-11-21
Replace which with command in Plexpy.py

wilmardo:
Replaces `which` with `command -v` to check for the python executable. 
Came across this since `which` is not available on a Centos 7 LXC container. Even with Python installed running Plexpy resulted in `Error: Python not found!` because of the missing which package.

Command is POSIX compatible (more info [see](https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script))
I [tested my changes](https://travis-ci.org/wilmardo/ansible-role-plex/builds/305442254) on the most popular platforms by adapting my Ansible role. It pulled my branch instead of the plexpy master and the tests complete nicely (which test that the webinterface is functioning).

### DIFF
--- a/PlexPy.py
+++ b/PlexPy.py
@@ -1,7 +1,7 @@
 #!/bin/sh
-''''which python    >/dev/null 2>&1 && exec python    "$0" "$@" # '''
-''''which python2   >/dev/null 2>&1 && exec python2   "$0" "$@" # '''
-''''which python2.7 >/dev/null 2>&1 && exec python2.7 "$0" "$@" # '''
+''''command -v python    >/dev/null 2>&1 && exec python    "$0" "$@" # '''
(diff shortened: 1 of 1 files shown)


JonnyWong16:
I've been wanting to get rid of this all together. It should really just use the shebang line.

Arcanemagus:
Note that some systems (such as FreeBSD) reserve `python` for Python 3, which is why that current mess is there (and why it can't use the `#!` without further trickery).

wilmardo:
Valid point, I personally see no problem in merging this. It makes Tautulli easier to run on different systems now and in the future.
I now need to install which before I can install Tautulli, not that much of a problem but fixed is fixed :)

heyhippari:
You could use the following as a shebang line to ensure it's using Python 2, even on systems where Python 3 is the default:

```
#!/usr/bin/env python2
```

wilmardo:
Superseeded by #1358

PR #1165, opened 2017-12-18
Fix safari-pinned-tab.svg and Lossless Compress

Fish2:
Lossless Compress all images which saves 304KB
Add Logo svg that was used to fix safari-pinned-tab.svg
Fix safari-pinned-tab.svg


JonnyWong16:
Thanks, but I don't want to compress all those files because they are a copy of the Plex assets.

I will take your svg files though.

Fish2:
@JonnyWong16 I can open a pull request with just the svg fix if you want but don’t really understand why you are against the compression as this is lossless so there is no loss in quality or loss of data that will impact plexpy and the upside is better load time when a lot of images are used on one page.

Fish2:
@JonnyWong16 Cheers For adding the svg and fixed svg

0cb147cd8f7cf7b6955f99c4e63ebd76927ac5b1
ad2f61132ad6588a51363ce5131df7e4b2e60211

JonnyWong16:
I put mentions in the changelog (see previous changelogs for examples).

And I'm removing the media flags from the info pages  (changing them to something else).

Fish2:
@JonnyWong16 ok cheers didn't know that sorry. 

Thats good news

PR #1181, opened 2017-12-29
Fix OpenSSL SSL.ConnectionType deprecation warning

### LINKED ISSUE #1142: pyOpenSSL "ConnectionType" deprecation warning
**Version:**
1.4.25

**Branch:**
master

**Commit hash:**
cfd30996264b7e9fe4ef87f02d1cc52d1ae8bfca

**Operating system:**
Windows Server 2016

**Python version:**
Python 2.7.14

**Problem:**
I wanted to enable SSL, so I did and got an error (can't recall the exact error) about not having OpenSSL available.
So I installed pyOpenSSL with pip. This worked fine.

But now my console window of PlexPy is getting the following warning:
`C:\Program Files (x86)\plexpy-1.4.22\lib/cherrypy\wsgiserver\ssl_pyopenssl.py:247: DeprecationWarning: ConnectionType has been deprecated, use Connection instead  if SSL and isinstance(sock, SSL.ConnectionType):`

Is that someting you can fix?
Thanks in advance!

felixbuenemann:
This patches the vendored cherrypy lib to work around a deprecation warning in pyOpenSSL 17.1.0 and later by checking and using the `SSL.Connection` class if available and falling back to `SSL.ConnectionType` otherwise.

Fixes #1142

### DIFF
--- a/lib/cherrypy/wsgiserver/ssl_pyopenssl.py
+++ b/lib/cherrypy/wsgiserver/ssl_pyopenssl.py
@@ -39,6 +39,10 @@ from cherrypy import wsgiserver
 try:
     from OpenSSL import SSL
     from OpenSSL import crypto
+    if hasattr(SSL, 'Connection'):
+        SSLConnectionType = SSL.Connection
+    else:
+        SSLConnectionType = SSL.ConnectionType
 except ImportError:
     SSL = None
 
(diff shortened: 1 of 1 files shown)


felixbuenemann:
Note that I chose the minimally invasive solution to patch the vendored cherrypy, because it has the lowest chance of breaking things.

In the long term cherrypy should be upgraded, the problematic code has been completely removed and replaced with [cheroot](https://github.com/cherrypy/cheroot) in cherrypy 9.0.0 and later.

However since the bundled cherrypy seems to be version 5.1.0 upgrading will likely require changes.

felixbuenemann:
Rebased due to history change on nightly branch.

@JonnyWong16 Is the nightly branch the proper target for pull request for v2?

JonnyWong16:
Thanks.

PR #1186, opened 2018-01-01
Make Tautulli logo clickable again

redfellow:
Clicking an `object` inside an `a` doesn't work, but clicking an `img` does. It's better to display a `.svg` file with `img` tag in any case.

https://caniuse.com/#feat=svg-img

### DIFF
--- a/data/interfaces/default/base.html
+++ b/data/interfaces/default/base.html
@@ -65,7 +65,7 @@
                     <span class="icon-bar"></span>
                 </button>
                 <a class="navbar-brand svg" href="home">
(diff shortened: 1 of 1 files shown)


JonnyWong16:
It is clickable. What browser are you using? Clear your browser cache?

redfellow:
Sorry, what a stupid mistake, I always have the network tab cache disable enabled, except for today. 👎 I can see the `::after` selector now that extends the clickable area.

You could also use `pointer-events: none;` on the `object` if you don't fancy changing the tag.

JonnyWong16:
Already did 19969a8

PR #1192, opened 2018-01-03
Attempt to send a notify even when subject or body is missing (Discord, Facebook, Slack)

redfellow:
PR example for https://github.com/JonnyWong16/plexpy/issues/1191


redfellow:
Just a note: when `Include Subject Line` is unticked, a filled out subject line is still required or a notification fails, because of this part in the code ` if not subject or not body: return` in `plexpy/notifiers.py`

You might want to band-aid that at some point. :)

JonnyWong16:
Thanks for letting me know.

PR #1195, opened 2018-01-04
Update date formats

Tommatheussen:
- Added correct Year date formats
- Replaced generic numeric values with actual examples
- Added some additional formats (mostly related to suffix, eg 1st, 2nd, ...)


JonnyWong16:
Thanks.

PR #1214, opened 2018-01-14
Update edit_user.html so it gives user option to undelete user.

mjervelund:
Added code so user has the option to undelete user thats marked as deleted. it's done so it only shows up if the user is marked as deleted.


JonnyWong16:
Added in 24ed63e.

Issue #1111, opened 2017-09-14
No IP for failed logins


briped:
**Version:**
VERSION 1.4.22

**Branch:**
Git Branch: master

**Commit hash:**
Git Commit Hash: ec685407bbd4924b7c94c2dab2f1754089e77322

**Operating system:**
Platform: Linux 4.10.0-33-generic (Lubuntu 17.04 x64)

**Python version:**
Python Version: 2.7.13 (default, Jan 19 2017, 14:48:08) [GCC 6.3.0 20170118]

**What you did?**
Attempted to find IP of failed logins in "/opt/plexpy/logs/plexpy.log" and the logs interface of the PlexPy UI.

I've tried to see a setting somewhere, or a FAQ somewhere, where I can enable this feature, but I have not been able to find it.

**What happened?**
No IPs logged for failed logins in either place. Only IP of successful logins shown in the logs interface of the PlexPy UI.

I know it says to link to the full log. If you still require more information I'll of course provide that.

I can't decide if it's a bug that this is missing, or if it's a feature request to add it. So I'll go for a bug for now.

`2017-09-15 00:02:56	DEBUG	PlexPy WebAuth :: Invalid login attempt from 'adminuser'.
2017-09-15 00:02:56	WARNING	PlexPy WebAuth :: Unable to retrieve Plex.tv user token for user 'adminuser'.
2017-09-15 00:02:56	WARNING	Failed to access uri endpoint /users/sign_in.xml. Status code 401
2017-09-14 23:59:14	DEBUG	PlexPy WebAuth :: Invalid login attempt from 'wronguser'.
2017-09-14 23:59:14	WARNING	PlexPy WebAuth :: Unable to retrieve Plex.tv user token for user 'wronguser'.
2017-09-14 23:59:14	WARNING	Failed to access uri endpoint /users/sign_in.xml. Status code 401
2017-09-14 23:52:00	DEBUG	PlexPy WebAuth :: Admin user 'adminuser' logged into PlexPy.
2017-09-14 23:39:00	DEBUG	PlexPy WebAuth :: Guest user 'plexuser' logged into PlexPy.`

**What you expected?**
To see the IP address of a failed login, for use in blocking mechanisms such as Fail2Ban.

**How can we reproduce your issue?**
Simply login to PlexPy using a wrong user and/or wrong password and check the logs.

JonnyWong16:
PlexPy does not log invalid logins.

briped:
[...]

PR #1242, opened 2018-02-05
Log IP address in web auth logs

jmt-gh:
This commit adds the IP address of the attempted user login to all applicable logs within web auth.

I also created a helper class to abstract some of the logic of extracting pieces of the HTTP request headers. This should help cut down on any future duplication.

As well, I modified a couple of the log messages to read a bit better with this new output

This resolves issue #1111

### DIFF
--- a/plexpy/webauth.py
+++ b/plexpy/webauth.py
@@ -103,7 +103,7 @@ def check_credentials(username, password, admin_login='0'):
         return True, u'guest'
     else:
         return False, None
-    
+
 def check_auth(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


Arcanemagus:
PR's should be done against the `nightly` branch, `dev` is still based on PlexPy not the Tautulli (v2) code (`beta`/`nightly`).

Also unfortunately after you did the work if I'm understanding it from the description what you have implemented here is already done in Tautulli (v2).

image

JonnyWong16:
Failed logins have already been added to v2.

image

Issue #753, opened 2016-06-14
{server_uptime} not reporting correctly


ronwbrown:
**Version:**
1.4.6
**Branch:**
Master
**Commit hash:**
d9474cdcc5e72ff311569a234e3abe7dec06d9a2
**Operating system:**
Server 2008
**Python version:**
latest
**What you did?**
Set notifications to post to facebook, including the {server_uptime} tag.
**What happened?**
messages posted shows hours of uptime, but only up to 24 hours, then resets to 0 again.
**What you expected?**
I expected to see "Days" or at least hours uptime above 24.
**How can we reproduce your issue?**
create a notification using the {server_uptime} tag.
if the server has been up for 25 hours, it will report 1 hour.

**What are your (relevant) settings?**
NA
**Link to logs:**
NA

nucshuco:
Same issue here on FreeNas 9.10.1

deed02392:
Same issue here on CentOS.

JonnyWong16:
`{server_uptime}` has been removed in v2 because there is no accurate way to get this info.

PR #1267, opened 2018-03-06
Fix {server_uptime} tag

deed02392:
Bug #753: part-fix. The `server.xml` endpoint returns two attributes:
```
createdAt="1497562639" updatedAt="1520351529"
```

Plexpy uses `updatedAt` for which it is unclear as to what event it references.

`createdAt` is _closer_ to the actual "server uptime" in that it seems to be the timestamp when the server was created. As of right now the `createdAt` value corresponds to roughly when I first installed Plex (from memory).

`{server_uptime}` is described in such a way it appears it should return what `uptime` returns when run on the console of the Plex server's host OS, but that value doesn't appear to be exposed anywhere by the Plex API that I can tell.

### DIFF
--- a/plexpy/notification_handler.py
+++ b/plexpy/notification_handler.py
@@ -550,8 +550,8 @@ def build_notify_text(session=None, timeline=None, notify_action=None, agent_id=
     server_times = plex_tv.get_server_times()
 
     if server_times:
-        updated_at = server_times['updated_at']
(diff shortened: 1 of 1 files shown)


JonnyWong16:
Since neither `updatedAt` or `createdAt` is an accurate "uptime", I have already removed the `{server_uptime}` parameter as an option in v2.

PR #1279, opened 2018-03-12
Added poster image support for IFTTT rich notifications

jsm174:
I primarily use IFTTT. Now that they support rich notifications, added poster image code found in the other notifiers.


jsm174:
Awesome! Thank you! Tested and working great.

PR #1282, opened 2018-04-09
Added the week number parameter

samip5:
I added the week number parameter for the newsletter. 😄


JonnyWong16:
1. Pull requests should be on `nightly`.
2. Tautulli uses the `arrow` library for dates.
3. The week number should be based on the start date of the newsletter, not the current date.
4. `week_day` should just be a newsletter parameter, not a class attribute.
```python
'week_day': self.start_date.isocalendar()[1]
```
5. Keep the parameters organized. Don't put `week_number` right in between `start_date` and `end_date`.
6. You also added `week_day` to the notification parameters, but you didn't add it to the list of available parameters in `common.py`.
7. Keep the date parameters grouped together, such as before `datestamp`.
8. Same thing, use the `arrow` library.
```python
'week_day': arrow.now().isocalendar()[1]
```

PR #1288, opened 2018-05-15
Fix Outlook not displaying any Pictures

Sonic-Y3k:
As Outlook is not displaying content of the newsletter with the css background-image tag correctly, we need to work around that with vml.

Note: Just a proposal, more work'll be needed.


JonnyWong16:
I'm not going to implement this "hack" into the templates.

You can add this support with your own custom templates.
https://github.com/Tautulli/Tautulli-Wiki/wiki/Frequently-Asked-Questions#newsletter-custom-template

PR #1289, opened 2018-05-19
Add Setting for X-Forwarded-Host Header

toddrob99:
Add setting for X-Forwarded-Host header, to fix issue with redirects
adding port to the URL when behind a reverse proxy (e.g.
https://tautulli.example.com ->
https://tautulli.example.com:8181/auth/login). If the reverse proxy is
running on nginx, this setting needs to be 'Host', and if Lighttpd, it
needs to be 'X-Host'.

See reports at
https://www.reddit.com/r/Tautulli/comments/7qajuz/tautulli_behind_reverse_proxy/


toddrob99:
I messed this up. Cloned beta instead of dev. Will redo.

PR #1290, opened 2018-05-19
Enhance reverse proxy support

toddrob99:
* Move `tools.proxy.on` setting from tree conf to options dict
because it was not present during redirects; use boolean value instead
of 0/1
* Create `redirect()` function to handle all redirects
* If `tools.proxy.on` is enabled in cherrypy config, check if
`X-Forwarded-Host` header is present (Apache). If not, fall back on
`X-Host` (lighttpd), `Origin` (Squid), or `Host` (nginx).

Tested using nginx reverse proxy on Synology (Host header). https://tautulli.example.com -> http://tautulli.local:8181 redirects as expected. Also tested custom port, https://tautulli.example.com:8443 -> http://tautulli.local:8181 also redirected as expected.

Note: I could not get a single `from plexpy.webserve import redirect`
statement to work at the top of `webauth.py`. I tried it all different
ways, and I could only get it to work by importing prior to each call to
`redirect()`. If you have any suggestions to make it work, I will be
happy to change it to a single import statement.


toddrob99:
Any chance of merging this in? I've been running with these changes and no issues since May.

toddrob99:
@JonnyWong16 can you please review and consider merging this pull request? Is there something else I can do to help?

JonnyWong16:
I do not like changing all the redirects.

I made a change to use the different header when resolving the proxy url. Let me know if this change works: 185099f1831249278344130cf2c2145ccb83c55a

toddrob99:
@JonnyWong16 I tested the current nightly branch and it works great. Thank you so much! 

I'm closing the pull request, although I was hoping to be listed as a contributor on the project by having a change merged in. If you happen to submit another commit related to the reverse proxy changes I suggested, maybe you could add me as a co-author. Thanks again.

PR #1291, opened 2018-05-20
Adding a simple login page notice

samip5:
This PR is about a notice to let users know that their information is securely checked against the API rather than having no indication of that anywhere.

Unfamiliar and users who have guest access will most likely appreciate it. Feel free to suggest edits if it should be done some other way, or way it's said.


samip5:
I messed it up, will redo.

PR #1293, opened 2018-05-20
Adding the notice/disclaimer to the login page

samip5:
I'm sorry for the PR's I messed up, but now it's not messed up. :smile: 

This PR is about a notice to let users know that their information is securely checked against the API rather than having no indication of that anywhere.

Unfamiliar and users who have guest access will most likely appreciate it. Feel free to suggest edits if it should be done some other way, or way it's said.

### DIFF
--- a/data/interfaces/default/login.html
+++ b/data/interfaces/default/login.html
@@ -58,6 +58,9 @@
                                 <input type="password" id="password" name="password" class="form-control">
                             </div>
                             <div class="form-footer">
+                              <div id="notice" class="alert" style="text-align: center; color: red;">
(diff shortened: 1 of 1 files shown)


Arcanemagus:
What does this notice actually accomplish?

Although Tautulli _doesn't_ store user's Plex credentials, there is nothing preventing someone from modifying the login page to do so.

samip5:
@Arcanemagus I don't necessarily know how useful that notice would be,  but this adds an notice to the login page regarding the authentication data being checked against the Plex API,  if I know correctly.

How does it then check Guest users login details if they are correct or not?

Is there a better way to inform users that it's safe to login with their details?

PR #1294, opened 2018-05-27
Add option to enable timestamps for rich metadata info on discord

samwiseg0:
Added the ability to send timestamps to discord.

### DIFF
--- a/plexpy/notifiers.py
+++ b/plexpy/notifiers.py
@@ -1149,8 +1149,6 @@ class DISCORD(Notifier):
                           'timestamp': helpers.utc_now_iso()
                           }
 
-            attachment['timestamp'] = helpers.utc_now_iso()
-
             if self.config['color']:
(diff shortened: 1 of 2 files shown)


JonnyWong16:
I don't think we need an option to include the timestamp. Just having it there all the time should be fine.

samwiseg0:
ok i can remove the option then.

plexpy/notifiers.py
```diff
@@ -1148,6 +1148,8 @@ def agent_notify(self, subject='', body='', action='', **kwargs):
             attachment = {'title': title
                           }
```
JonnyWong16: Just put the timestamp right inside the `attachment`.

samwiseg0: ok

JonnyWong16:
Squash and it's good.

samwiseg0:
I think it should be good now?

JonnyWong16:
Um... I want one single commit.
https://gist.github.com/patik/b8a9dc5cd356f9f6f980#combining-the-commits

PR #1295, opened 2018-05-27
Add timestamps for rich metadata info on discord

samwiseg0:
Sorry I could not for the life of me get git to behave and make it one commit. This should be right.


JonnyWong16:
Thanks.

PR #1301, opened 2018-08-16
Refactor OAuth Login

samwiseg0:
I was able to refactor the OAuth login to work by using `get_plextv_resources()` vs `get_plextv_server_list()`. I think this is a better method of determining the client accessToken in general since it follows the same method as the plex web app/players. With these changes OAuth should work with or without remote access just like it does for the plex web app and devices.


JonnyWong16:
Thanks.

PR #1302, opened 2018-08-17
Very minor formatting change in newsletter

simonk83:
Add a space after the slash between shows and episodes (it annoys me every time I see it)

eg:  Was 10 shows /100 episodes
       Now 10 shows / 100 episodes


JonnyWong16:
Pull requests should be made to the `nightly` branch.

Please read the guidelines here:
https://github.com/Tautulli/Tautulli/blob/master/CONTRIBUTING.md

Also note: These changes won't work 100% because all newlines/spaces are stripped when the email is sent to save bytes.

simonk83:
Ah ok, apologies for the incorrect branch.  I won't bother resubmitting then but you get the idea ;)

The newsletter looks pretty squished and weird on mobile as well, just a heads up.

PR #1308, opened 2018-09-10
Fix typo in systemd group value

ldumont:
New PR to fix https://github.com/Tautulli/Tautulli-Issues/issues/119


JonnyWong16:
Thanks.

PR #1311, opened 2018-09-15
Add chown instructions per major distros

samwiseg0:


samwiseg0:
I need to add another command for Ubuntu :(

PR #1314, opened 2018-09-19
Fixed typo

soehlert:
There was a typo on the group name in the systemd unit file. This was keeping tautulli from starting up correctly.


Arcanemagus:
This is already fixed on the `nightly` branch:
https://github.com/Tautulli/Tautulli/blob/67c333e86ee491504a222064af0c273c8b370d52/init-scripts/init.systemd#L54

soehlert:
Apologies, didn't see that.

PR #1316, opened 2018-09-19
Fix the type of the Content Rating notification parameter

Arcanemagus:
The "Content Rating" notification parameter was incorrectly marked as an integer, leading to all values being cast to the number 0. This made it so every single content rating was the same value in conditions.

Fixes https://github.com/Tautulli/Tautulli-Issues/issues/122.


JonnyWong16:
Thanks.

PR #1318, opened 2018-09-21
1 improvement & 1 bug fix regarding view offset in activity handler files

Sheigutn:
1. This change in activity_handler.py results in a more up-to-date watch percentage for the currently playing items on the history page. This is done by updating the view offset field whenever it was changed in the session state notification object.

2. It happened to me several times that short songs showed up with a watch percentage of 0%. This is because the "view_offset" field in the "raw_stream_info" json was not always as up-to-date as the one in the session object. By excluding "view_offset" from being overwritten, the correct value will be written to "session_history."

Example of what I am talking about:
Before the fix:
chrome_2018-09-21_02-42-33
After the fix:
chrome_2018-09-21_02-42-52

### DIFF
--- a/plexpy/activity_handler.py
+++ b/plexpy/activity_handler.py
@@ -228,6 +228,7 @@ class ActivityHandler(object):
             this_state = self.timeline['state']
             this_rating_key = str(self.timeline['ratingKey'])
             this_key = self.timeline['key']
+            this_view_offset = self.timeline['viewOffset']
 
             # Get the live tv session uuid
             this_live_uuid = this_key.split('/')[-1] if this_key.startswith('/livetv/sessions') else None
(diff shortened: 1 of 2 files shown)


plexpy/activity_handler.py
```diff
@@ -251,6 +253,10 @@ def process(self):
                         if int(time.time()) - db_session['stopped'] > 60:
                             self.update_db_session()
 
+                    # Update db session when view offset changes
+                    if this_view_offset != last_view_offset:
+                        self.update_db_session()
```
JonnyWong16: I don't like updating the database for every single websocket event. This is why I have the line right above to update every 60 seconds. I think it is acceptable for the currently active items in the history table to be at most 60 seconds out. They will be correct when the stream is stopped anyways.

JonnyWong16:
I am okay with change number 2 for excluding `view_offset` from being overwritten.

JonnyWong16 (changes requested):

plexpy/activity_handler.py
Sheigutn: Oh, okay! I'll revert this change then.

JonnyWong16 (approved):

JonnyWong16:
Thanks.

PR #1320, opened 2018-09-26
Telegram api url as variable

snvad:
In some cases, the api url is different from the main one. Added the ability to change this parameter.


snvad:
If the address is blocked by the Internet provider or at the workplace (If necessary, use a mirror)

JonnyWong16:
Sorry but that issue case is not Tautulli's concern. You will have to find your own solution.

PR #1327, opened 2018-10-18
Added new helpers for time, and added the 24 hour check to avoid notification spam

samip5:
Any thoughts on this?


samwiseg0:
Wouldn't setting the check interval to 24hrs accomplish the same thing? 

What benefit is this change over changing the interval?

JonnyWong16:
`CHECK_GITHUB_INTERVAL` is already a config value.

PR #1334, opened 2018-11-05
init.systemd add restart policy

zetneteork:
Improve systemd unit file to restart service automatically.

```bash
Restart=always
RestartSec=5
```

### DIFF
--- a/init-scripts/init.systemd
+++ b/init-scripts/init.systemd
@@ -53,6 +53,8 @@ GuessMainPID=no
 Type=forking
 User=tautulli
 Group=tautulli
+Restart=always
+RestartSec=5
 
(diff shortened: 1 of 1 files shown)


Arcanemagus:
Having this be the default seemed too risky since quite often when first setting things up users will run into a scenario where it will crash right away due to a misconfiguration. In #1341 I've added this recommendation as a note along with the other suggested enhancements.

Note that `on-abnormal` is the policy I used there as it seems a better fit (`always` would prevent Tautulli from being able to shut itself down).

Thanks for submitting a PR though as this is a good idea!

JonnyWong16:
#1341

PR #1335, opened 2018-11-28
Added proxy config for Telegram notifier

alexxoid:
Added ability to enter proxy settings for Telegram notifications if "api.telegram.org" is blocked in some situations.


JonnyWong16:
#1320

alexxoid:
Why?
It's just the ability to make notifier config more flexible. Just like we can get own port for Plex. People can use proxy in local networks. Without direct access to Plex Server Telegram via proxy is god way to know some statistics.

samwiseg0:
You can proxy at the OS level if that is what you need.

PR #1337, opened 2018-12-30
Fix css min-width on mobile screen

n-peugnet:
I removed the min-width property on small screen to remove horizontal scrollbars


n-peugnet:
min-width-350 min-width-auto

n-peugnet:
Hey, any news about merging this @JonnyWong16

PR #1339, opened 2019-02-09
ADD Basic support of Synology Chat Webhooks for Notification

axelbd:
Hello,

I've added Synology Chat as a notification agent. Support is now minimal but if i've time i'll try to add PrettyMetadata.

Regards,

### DIFF
--- a/plexpy/notifiers.py
+++ b/plexpy/notifiers.py
@@ -92,7 +92,8 @@ AGENT_IDS = {'growl': 0,
              'groupme': 22,
              'mqtt': 23,
              'zapier': 24,
-             'webhook': 25
+             'webhook': 25,
+             'synochat': 26
              }
 
 DEFAULT_CUSTOM_CONDITIONS = [{'parameter': '', 'operator': '', 'value': ''}]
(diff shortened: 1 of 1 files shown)


JonnyWong16:
Why is this necessary? Just use the standard Webhook notifier.

axelbd:
I didn't managed to make it work with the standard Webhook notifier as Chat require the "payload" parameter in the url and the content-type application/x-www-form-urlencoded instead of application/json.

Admitting that somebody managed to do it with the standard Webhook notifier, using a separate notifier simplify the formatting of the notification text by removing the need to write the notification messages in JSON format in the Data tab of the notifier settings.

Do you think this request could be usefull ? If you have a working example of Chat integration with the standard notifier i would gladly use it as i don't really know how many users of tautulli need this specefic notifier.

JonnyWong16:
Then the better pull request is to add the option of changing the `Content-Type` header for the Webhook notifier.

axelbd:
I'll try and test but i doubt this will suffice as the `payload` parameter is another thing that might need to be implemented. Thanks for the help :)

Arcanemagus:
> as Chat require the "payload" parameter in the url and the content-type application/x-www-form-urlencoded instead of application/json.

Note that this _isn't_ a webhook, you should contact Synology to either implement proper webhooks or stop calling them that.

PR #1341, opened 2019-02-23
Systemd restart policy note

Arcanemagus:
If Tautulli ever crashes due to a failure of some sort the policies mentioned here will automatically restart it, with the caveat that they will _always_ restart it, even if it is going to crash right away again!

`on-abnormal` appears to be the best policy choice here from what [the documentation](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=) says.

Closes #1334.


vertig0ne:
You can use `StartLimitBurst` and `StartLimitIntervalSec` to catch repeated immediate errors and then halt the process from starting further

EDIT:

Example:
StartLimitInterval=90
StartLimitBurst=3

If it restarts 3 times in a 90 second period. It will be considered failed.

JonnyWong16:
Thanks.

PR #1342, opened 2019-03-02
Add gzip encoding as a configurable option

mew1033:
If you reverse proxy Tautulli through IIS, you have to do funky stuff with the encoding to make it work. This just adds an option to disable it inside the app to fix the problem.

I'm only _mostly_ sure I added this config option correctly. Let me know if I did something dumb.

### DIFF
--- a/Tautulli.py
+++ b/Tautulli.py
@@ -240,6 +240,7 @@ def main():
         'http_root': plexpy.CONFIG.HTTP_ROOT,
         'http_environment': plexpy.CONFIG.HTTP_ENVIRONMENT,
         'http_proxy': plexpy.CONFIG.HTTP_PROXY,
+        'http_gzip': plexpy.CONFIG.HTTP_GZIP,
         'enable_https': plexpy.CONFIG.ENABLE_HTTPS,
         'https_cert': plexpy.CONFIG.HTTPS_CERT,
         'https_cert_chain': plexpy.CONFIG.HTTPS_CERT_CHAIN,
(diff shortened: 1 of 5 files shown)


JonnyWong16:
`gzip` is a web standard and should not need to be disabled. The issue is likely in your IIS configuration.

mew1033:
@JonnyWong16 I agree, it shouldn't need to be disabled. But why can't it be a configurable option?
Note that IIS still has gzip encoding when connecting to remote clients. This change is only needed when IIS is acting as a reverse proxy, and it only changes the encoding on the connections from IIS to Tautulli.

Arcanemagus:
You should look into what is broken in your IIS configuration that this is required, instead of submitting a PR to other apps in order to **break** them to work with IIS.

mew1033:
@Arcanemagus I know exactly what's going on with IIS. It's not a great reverse proxy solution, and it doesn't fully support gzip encoding to the servers/services it's fronting. See [here](https://blogs.msdn.microsoft.com/friis/2016/08/25/iis-with-url-rewrite-as-a-reverse-proxy-part-2-dealing-with-500-52-status-codes/) for an example.

This change isn't _required_ to get Tautulli working behind IIS, there are other ways about it. However, it does provide an option for Tautulli to deal with poorly configured reverse proxies.

Why wouldn't you want your software to be _more_ configurable? _More_ supporting of less-than-ideal hosting situations?

samwiseg0:
> Why wouldn't you want your software to be more configurable? More supporting of less-than-ideal hosting situations?

That is not an approach we want to take. If we use that logic settings would be riddled with options that most people don't need. Furthermore, having all those options just opens us up for many more support issues and increases our testing surface area.

I personally would not want to use a product that cannot support basic gzip. Use a real reverse proxy. That is the solution. If your hosting solution forces you to use IIS. Find a new hosting solution. That is a much simpler approach.

PR #1344, opened 2019-03-09
Update init.systemd

Jabcob:
Edited user creation and directory ownership instructions in the configuration notes for clarity. Current version may give novice users the impression that the ownership command is only executed on CentOS/Fedora.


JonnyWong16:
Thanks.

PR #1345, opened 2019-03-11
Allow time_queries to be passed into get_watch_time_stats methods as arguments.

idlegravity:
Modified Users.get_watch_time_stats and Libraries.get_watch_time_stats to take an arbitrary list of time_queries rather than using the hard-coded list [1, 7, 30, 0]. Callers of these methods have been updated to use the previously hard-coded list as defaults for backwards compatibility.

The motivation for this change is to allow custom scripts greater flexibility in querying watch time stats (e.g. querying watch time for the current calendar week, month or year).

### DIFF
--- a/plexpy/libraries.py
+++ b/plexpy/libraries.py
@@ -759,7 +759,7 @@ class Libraries(object):
                 # If there is no library data we must return something
                 return default_return
 
-    def get_watch_time_stats(self, section_id=None, grouping=None):
+    def get_watch_time_stats(self, section_id=None, grouping=None, time_queries=None):
         if not session.allow_session_library(section_id):
             return []
 
@@ -768,7 +768,6 @@ class Libraries(object):
 
         monitor_db = database.MonitorDatabase()
 
-        time_queries = [1, 7, 30, 0]
         library_watch_time_stats = []
 
         group_by = 'session_history.reference_id' if grouping else 'session_history.id'
--- a/plexpy/users.py
+++ b/plexpy/users.py
@@ -415,7 +415,7 @@ class Users(object):
                 # Use "Local" user to retain compatibility with PlexWatch database value
                 return default_return
 
-    def get_watch_time_stats(self, user_id=None, grouping=None):
+    def get_watch_time_stats(self, user_id=None, grouping=None, time_queries=None):
         if not session.allow_session_user(user_id):
             return []
 
@@ -424,7 +424,6 @@ class Users(object):
 
         monitor_db = database.MonitorDatabase()
 
-        time_queries = [1, 7, 30, 0]
         user_watch_time_stats = []
 
         group_by = 'reference_id' if grouping else 'id'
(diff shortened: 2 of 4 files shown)


plexpy/libraries.py
```diff
@@ -768,7 +768,6 @@ def get_watch_time_stats(self, section_id=None, grouping=None):
 
         monitor_db = database.MonitorDatabase()
 
-        time_queries = [1, 7, 30, 0]
```
JonnyWong16, line 771: The default `time_queries` should be added here otherwise a call to `get_watch_time_stats` without any `time_queries` will fail below.

```
if time_queries is None:
    time_querires = [1, 7, 30, 0]
```

plexpy/users.py
```diff
@@ -424,7 +424,6 @@ def get_watch_time_stats(self, user_id=None, grouping=None):
 
         monitor_db = database.MonitorDatabase()
 
-        time_queries = [1, 7, 30, 0]
```
JonnyWong16, line 427: The default `time_queries` should be added here otherwise a call to `get_watch_time_stats` without any `time_queries` will fail below.

```
if time_queries is None:
    time_querires = [1, 7, 30, 0]
```

plexpy/webserve.py
```diff
@@ -552,7 +552,7 @@ def library_watch_time_stats(self, section_id=None, **kwargs):
 
         if section_id:
             library_data = libraries.Libraries()
-            result = library_data.get_watch_time_stats(section_id=section_id)
+            result = library_data.get_watch_time_stats(section_id=section_id, time_queries=[1, 7, 30, 0])
```
JonnyWong16, line 555: Can leave out the `time_queries` argument since the default is added to `get_watch_time_stats`.

plexpy/webserve.py
```diff
@@ -1183,7 +1190,7 @@ def user_watch_time_stats(self, user=None, user_id=None, **kwargs):
 
         if user_id or user:
             user_data = users.Users()
-            result = user_data.get_watch_time_stats(user_id=user_id)
+            result = user_data.get_watch_time_stats(user_id=user_id, time_queries=[1, 7, 30, 0])
```
JonnyWong16, line 1193: Can leave out the `time_queries` argument since the default is added to `get_watch_time_stats`.

plexpy/webserve.py
```diff
@@ -816,11 +817,17 @@ def get_library_watch_time_stats(self, section_id=None, grouping=None, **kwargs)
                      ]
             ```
         """
+        if query_days:
+            query_days = map(int, query_days.split(","))
+        else:
+            # Default values for backwards compatibility
+            query_days = [1, 7, 30, 0]
```
JonnyWong16, line 824: Can leave out the `else` block since the default is added to `get_watch_time_stats`.

plexpy/webserve.py
```diff
@@ -1431,11 +1439,17 @@ def get_user_watch_time_stats(self, user_id=None, grouping=None, **kwargs):
...
```
JonnyWong16, line 1446: Can leave out the `else` block since the default is added to `get_watch_time_stats`.

JonnyWong16 (commented):
Just need to move the default `time_queries` inside the `get_watch_time_stats` function.

PR #1349, opened 2019-05-04
Updated info.html so that the star rating and tooltip matches

karl-sjogren:
The rating value was divided by 2 when calculating the 5 star value but not when writing it to the tooltip. So the tooltip said 7 and 4 out of 5 stars were lit. This change updates the tooltip to also divide by two and also changes the docs at the top of the file to say that the input is 1 to 10.


JonnyWong16:
This isn't a bug. Ratings in Plex are out of 10, but I don't want to show 10 stars in the UI so it is rounded out of 5 for the stars. This is the same way Plex shows star ratings in their UI.

image

PR #1355, opened 2019-11-03
Update tautulli.css

zobe123:
better look if "(hw)" tag is used 2 times in video line


JonnyWong16:
I do not like the smaller padding.

zobe123:
image

maybe there is any prettier solution for this...

JonnyWong16:
The smaller padding doesn't even work for 1080p.

image

PR #1356, opened 2019-11-03
Move vendor libs to standard requirements.txt

fliiiix:
This addresses this issue: https://github.com/Tautulli/Tautulli-Issues/issues/200

Which then will allow to upgrade the codebase to python3. 

Help is appreciated to move everything.


mercury00:
Ah - I have a list of all the vendor libs. Note that there are three libraries that are not available from requirements.py as they're not in pypi. I'll update a branch with the edits.

mercury00:
Ok, I've finished this with request #1365.

I think. Someone will need to check.

fliiiix:
@JonnyWong16 care to clarify why?

JonnyWong16:
Superceded by #1365.

PR #1357, opened 2019-11-10
Add grafana notifier

hdkiller:
This patch allows tautulli to post annotations to grafana.

https://grafana.com/docs/http_api/annotations/


JonnyWong16:
Thanks, but this should just use the Webhook notification agent.

I just added the option for custom Webhook headers in ddbd4865, so it should work now.

PR #1359, opened 2019-12-03
Fix startup if the python executable is already v3+

tamisoft:
Problem surfaced on Fedora 31, where the default
/usr/bin/python link to point to the Python 3 interpreter.

This fix that by evaluating the version using python itself.


binyaminyblatt:
this guy did it 
https://github.com/zSeriesGuy/Tautulli/
With multiple PMS servers.
and other Features

JonnyWong16:
Shebang line updated in v2.5.

PR #1361, opened 2019-12-20
add summary_short params

Akior:
Want to use summary params for notification agent
Get telergam limitation (1024 symbols) for caption message because summary is oversize
Create short summary block

This method cut text and can cut word for N symbols. 
Maybe there are another method of deal with it, but Im not high skill with python.

### DIFF
--- a/plexpy/common.py
+++ b/plexpy/common.py
@@ -441,6 +441,7 @@ NOTIFICATION_PARAMETERS = [
              {'name': 'Labels', 'type': 'str', 'value': 'labels', 'description': 'A list of labels for the item.'},
              {'name': 'Collections', 'type': 'str', 'value': 'collections', 'description': 'A list of collections for the item.'},
              {'name': 'Summary', 'type': 'str', 'value': 'summary', 'description': 'A short plot summary for the item.'},
+             {'name': 'Summary_short', 'type': 'str', 'value': 'summary_short', 'description': 'A very short plot summary for the item.'},
              {'name': 'Tagline', 'type': 'str', 'value': 'tagline', 'description': 'A tagline for the media item.'},
              {'name': 'Rating', 'type': 'float', 'value': 'rating', 'description': 'The rating (out of 10) for the item.'},
              {'name': 'Critic Rating', 'type': 'int', 'value': 'critic_rating', 'description': 'The critic rating (%) for the item.', 'help_text': 'Ratings source must be Rotten Tomatoes for the Plex Movie agent'},
--- a/plexpy/common.py
+++ b/plexpy/common.py
@@ -441,7 +441,6 @@ NOTIFICATION_PARAMETERS = [
              {'name': 'Labels', 'type': 'str', 'value': 'labels', 'description': 'A list of labels for the item.'},
              {'name': 'Collections', 'type': 'str', 'value': 'collections', 'description': 'A list of collections for the item.'},
              {'name': 'Summary', 'type': 'str', 'value': 'summary', 'description': 'A short plot summary for the item.'},
-             {'name': 'Summary_short', 'type': 'str', 'value': 'summary_short', 'description': 'A very short plot summary for the item.'},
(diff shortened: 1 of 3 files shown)


Arcanemagus:
You need to change this to target the `nightly` branch, as explained in [CONTRIBUTING](https://github.com/Tautulli/Tautulli/blob/master/CONTRIBUTING.md).

JonnyWong16:
I would much rather limit the text to 1024 characters for the Telegram agent specifically, instead of some arbitrary 400 characters.

See 688d28b5ea2360769bd58504fc8fb905521ce676 on how Discord is limited to 2048 characters.

Akior:
> I would much rather limit the text to 1024 characters for the Telegram agent specifically, instead of some arbitrary 400 characters.
> 
> See [688d28b](https://github.com/Tautulli/Tautulli/commit/688d28b5ea2360769bd58504fc8fb905521ce676) on how Discord is limited to 2048 characters.

The limit of telegram is 1024 characters, but sending message for English letters is 

     sendPhoto -caption "EnglishText" 
and for Russian example:

     sendPhoto -caption "\u0420\u0443\u0441\u0441\u043a\u0438\u0439\u0442\u0435\u0441\u0442"
Tau have code:

     if len(text) > 1024:
                    data['disable_notification'] = True
So for the same limit char we have not same length of text.
If I do something like this:

     if len(text) > 2048:
                    data['disable_notification'] = True
     else:
                    data['caption'] = text[:1745] + (text[1745:] and '...')
English text will be drop if char more than 1024, but Russian will work. English text error:

     Request response: {"ok":false,"error_code":400,"description":"Bad Request: MEDIA_CAPTION_TOO_LONG"}

Anyway, the biggest text is "summary", so limit to summary_short will work for all Languages

JonnyWong16:
The issue isn't with `/sendPhoto`. The caption is already limited to 1024 characters. You don't need to change that.

The issue is the 4096 character limit for the `/sendMessage`.

https://github.com/Tautulli/Tautulli/blob/master/plexpy/notifiers.py#L3453

```
data['text'] = text[:4093] + (text[4093:] and '...')
```

Akior:
> The issue isn't with `/sendPhoto`. The caption is already limited to 1024 characters. You don't need to change that.
> 
> The issue is the 4096 character limit for the `/sendMessage`.
> 
> https://github.com/Tautulli/Tautulli/blob/master/plexpy/notifiers.py#L3453
> 
> ```
> data['text'] = text[:4093] + (text[4093:] and '...')
> ```

I mean limitation of caption works bad. I can send more RU text with fix > 2048. 
Looks like len(text) != len(limitation) because unicode

If you don't want to add summary_short, you can close request.
But I will be much appreciated, If you can add some fix for sendPhoto with caption for Ru(or maybe any non English lang) segment

JonnyWong16:
Text just needs to be decoded to Unicode.

https://github.com/Tautulli/Tautulli/blob/master/plexpy/notifiers.py#L3442

```
if len(text.decode('utf-8')) > 1024:
```

Akior:
> if len(text.decode('utf-8')) > 1024:

                if len(text.decode('utf-8')) > 1024:
                    data['disable_notification'] = True
                else:
                    data['caption'] = text.decode('utf-8')[:1021] + (text.decode('utf-8')[1021:] and '...')

But we don't need 1024 char check anymore. 

Can you add this to tau main ?

Akior:
Your solution doesn't split text of caption. It split telegram message:
Снимок экрана от 2020-01-21 13-36-13
Without 1024 check and :

     data['caption'] = text.decode('utf-8')[:1021] + (text.decode('utf-8')[1021:] and '...')

Снимок экрана от 2020-01-21 13-38-49

JonnyWong16:
That is correct behavior. Less than 1024 is a caption, more than 1024 is a separate message.

Akior:
> That is correct behavior. Less than 1024 is a caption, more than 1024 is a separate message.

Can you add "force caption" option with 1024 splitting?

JonnyWong16:
No.

PR #1362, opened 2020-01-19
Slightly reduce margin on stat cards

RafaelSchridi:
I've got a single vertical 1080p monitor and currently all the cards are in a single line, but reducing the margins on cards slightly allows them to appear in dual lines.

https://i.imgur.com/FyNBVtv.png Is how it currently looks.

### DIFF
--- a/data/interfaces/default/css/tautulli.css
+++ b/data/interfaces/default/css/tautulli.css
@@ -1159,7 +1159,7 @@ a .dashboard-activity-metadata-user-thumb:hover {
     height: 160px;
     min-width: 350px;
     max-width: 500px;
-    margin-right: 25px;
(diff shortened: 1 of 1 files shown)


JonnyWong16:
The reason that the margin is set to `25px` is so that the cards line up with the recently added posters. It will bother me like crazy if they don't line up.





If you want to change the margin, I suggest you just install the [Stylus browser add-on](https://add0n.com/stylus.html) and change the css for yourself.

```
.dashboard-stats-instance {
    margin-right: 15px !important;
}
```

Also, pull requests should be made to the `nightly` branch. Please read [CONTRIBUTING.md](https://github.com/Tautulli/Tautulli/blob/master/CONTRIBUTING.md).

PR #1363, opened 2020-01-23
Change humanFileSize to default to SI notation

aaronldunlap:
The current code creates output like “12.0 GiB” instead of much more readable “12 GB”. This change modifies the default behavior of humanFileSize() to do just that.


JonnyWong16:
Thanks.

PR #1364, opened 2020-02-14
Add crossorigin use-credentials attribute to manifest tag

pkoenig10:
When serving Tautulli behind an authentication proxy (ex. https://github.com/pusher/oauth2_proxy) the manifest request will fail because cookies are not sent with the manifest request.

See https://developer.mozilla.org/en-US/docs/Web/Manifest:
> Note: If the manifest requires credentials to fetch - the `crossorigin` attribute must be set to `"use-credentials"`, even if the manifest file is in the same origin as the current page.

This PR adds the necessary attribute so that cookies will be sent with manifest requests.

### DIFF
--- a/data/interfaces/default/base.html
+++ b/data/interfaces/default/base.html
@@ -28,7 +28,7 @@
 
     <!-- ICONS -->
     <!-- Android -->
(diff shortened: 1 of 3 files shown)


JonnyWong16:
Thanks.

PR #1365, opened 2020-02-29
Feature/move libs

mercury00:
Finished fliiiix's removing of vendor libs from libs folder and creating requirements.txt file for pip instead.

Installing required libraries is first done by a 'pip install -r requirements.txt' prior to running Tautulli.


fliiiix:
:tada: very cool I try to checkout the changes this week

fliiiix:
Sorry took me longer than excepted:
here are a few things:

there is a `=` sign missing 
```
-pytz=2019.3
+pytz==2019.3
```

I guess pyOpenSSL is missing from `requirements.txt`
```
2020-04-06 19:20:34 - WARNING :: MainThread : The pyOpenSSL module is missing. Install this module to enable HTTPS. HTTPS will be disabled.
``` 

With my existing config the login didn't work failed with a 500 error code somewhere.
I also tried to create a new config which is always redirecting to the setup screen forever. 

so this is definitely not ready yet

mercury00:
Great, I'll check it out.

Are you on Windows or Linux?

I've only tested with Linux so far.

I had the redirect problem earlier, it was fixed in the html file rather than the python code. I'll try a clean pull and see what I get.

--
Aaron

________________________________

mercury00:
Ok - I updated the requirements.txt. Added the missing pyopenssl module (it was previously hidden in urllib3) and the missing idna and cryptography libs.

I did a clean install, created a virtualenv for tautulli, followed the install instructions plus 'pip install -r requirements.txt', walked through the first time setup, and everything came up just fine. I haven't had any errors yet.

I've replaced the pbkd2 library with the original one; it's less secure I'm sure, but this way your old password should work from a previous install. My assumption is that replacing the static pbkdf2 with the version from pip causes it to use a different hashing method somewhere - if so, using the original version should work ok.

Let me know how your testing goes.

fliiiix:
Looks very promising. Worked for my existing config. And worked for a new setup. 
So I looked at the commits would be cool if you can rebase it on the latest master. 

And I can tests it on windows and FreeBSD.

mercury00:
Ah ok. I tried to rebase yesterday - I think I did it wrong.

I'll try again; do it right this time.

--
Aaron

________________________________

JonnyWong16:
Please change the pull request to the `nightly` branch.

But just an FYI, I have updated a lot of the libraries (https://github.com/Tautulli/Tautulli/compare/nightly...python3), so there will be a lot of rework for the PR.

mercury00:
Ok. I will update the requirements.txt file to match nightly.

--
Aaron

________________________________

JonnyWong16:
My library updates are not on `nightly` yet.

mercury00:
Ah, I see, python3.

For python3 and if this pr eventually works, it will mostly be a matter of keeping the requirements.txt file updated.

However, for this to be in production, we need to add a method to the updater to run the install of the packages in requirements.txt after it's updated itself to the latest master branch etc.

I have a script that can do that, but updates will certainly break if tautulli no longer has the right libs did match the updated code.

Anyway, I may supercede this pr by branching from nightly then committing each lib change one by one, then pushing only when they're all done. Will take about 20m now that I have the end product. I think it will be easier to follow what I've changed (just removing libs and adding their pypi versions in requirements.txt), otherwise I have big commits with massive changes to trawl through to make sure I didn't break something in this PR.

--
Aaron

________________________________

samwiseg0:
It would be ideal to have a proper installer vs a script for windows installs.

mercury00:
Ah, sure - on linux, when one runs ./Tautulli.py I believe there's an updater in checkversion, what I was thinking was doing something along the lines of (but sanely with proper handling of errors etc):

```
from pip._internal.cli.main import main as pip_main
with open('requirements.txt', 'r') as packages:
    for package_name in packages:
        pip_main(['install', '--target={:<}'.format(lib_path), '--upgrade', package_name])
```

lib_path could still be `./lib` , perhaps for instance if there were something placed in `.gitignore`.

At the moment there's a chicken-and-egg problem that first-time installs need to overcome with something like `pip install --target=./lib --requirement requirements.txt` or such as tautulli user (the tautulli user needs permission to write the lib files one way or another, so probably `su -s /bin/bash tautulli` or something).
Anyway, something to think about in parallel with the requirements.txt file replacing committed libraries.

professionaltart:
@mercury00 I can try testing this on windows 10 if you'd like.  I pulled the python3 branch thinking it was what you were working on, and then saw no requirements.txt - but then I didn't fine one on your fork, either.

mercury00:
I'll update this with the latest nightly.

The requirements.txt is on my fork but in the branch called feature/move-libs.

I'll update this though, and that branch will be called feature-nightly/move-libs - later this week.

JonnyWong16:
It's also a bad idea for the updater to install (global) packages for the user. You don't know what other software they might be running and having Tautulli update a package might break that other software. Most people don't have a Python virtual environment setup to run Tautulli.

In order for this to work, Tautulli needs to move away from having users running from source. It should move towards having installers available for as many platforms as possible, such as the ones for Windows/macOS or Docker. This way the user does not need to install any requirements (or possibly even need to install Python). The requirements file would only be used for development purposes.

mercury00:
The updater wouldn't install global files. Tautullis libs folder is where everything is installed. It's an easy flag to add to the script that would do the updates. Pip will allow you to install files anywhere.

An installer is a good idea, sure, but just to make it clear, users would not be running a command to install requirements themselves, they would tell tautulli to do it.

--
Aaron
________________________________

grawlinson:
Any status on this? Just curious as to why this seems abandoned.

samwiseg0:
> Any status on this? Just curious as to why this seems abandoned.

This is closed. If you have any further questions please use our discord channel.

PR #1366, opened 2020-03-01
Fix display of IPv6 addresses in activity panes

felixbuenemann:
* Fix wrong width of ip-address field, causing it to be displayed as "…", when exceeding 125px
* Add tooltip with full IP address on hover, if the address is longer than 20 characters (maximum length that can be shown in 125px)


JonnyWong16:
Thanks.

PR #1367, opened 2020-03-09
Start date parameter fix

KaasKop97:
Originally start_date would only return data if you entered the current date since it used the equals operator instead of the greater than operator.
Added a clause where if it matches the date string in regex it then uses the greater than operator.

P.S.: I apologize in advance if I've done something wrong. It's my first time doing this :)


JonnyWong16:
`start_date` is supposed to be a specific day. It is used for retrieving history for the specific date when you click a data point on the Daily Play count graph.

PR #1373, opened 2020-05-09
Specify a full path to python

Arcanemagus:
Apparently some systemd implementations (all?) **require** a full path, so change `python` to `/bin/python` which should be viable on most systems.


JonnyWong16:
I think removing the interpreter and using the shebang line is safer.

PR #1374, opened 2020-07-11
Add first_seen to user IP Table

dotsam:
This PR adds a first_seen column to the user IP detail table, and also sets a title attr on both the first_seen and last_seen columns that shows the full date/time in the user's set formats.


JonnyWong16:
Cool, thanks.

PR #1375, opened 2020-07-15
Snap Package integration

capruro:
Hello,

I built the snap package and I'm manually maintaining, recompiling the file every release, you can find it on: https://github.com/capruro/tautulli-snap
It would be great if you pull the snapcraft into your master branch so that every commit will trigger a new snap automatically! :)
This will sync the release with all the users and avoid manual job. Hundreds of users are already using it.
Let me know.


Arcanemagus:
PR's should be against the `nightly` branch (as described in [CONTRIBUTING](https://github.com/Tautulli/Tautulli/blob/master/CONTRIBUTING.md)).

You should also follow your own Editorconfig you are trying to add 🤣:
image

Turux:
Surely bragging ain't the way forward... @capruro made a working snap package in his free time that is contributing to the success of your app. A thank you could have been a good start...

capruro:
@Arcanemagus: As you would, I change the pull request to nightly branch.
Removed the Editorconfig as purely for my purpose :)

JonnyWong16:
Thanks for your pull request but here are my concerns:

1. By adding this to the Tautulli repository, this make the snap package "official" which means I have to then provide support for it. I have no experience with snap packages and neither does anyone else on the team. Unfortunately, we cannot provide support for something that we do not use ourselves or understand completely. This would need to be implemented in a complete GitHub workflow, similar to the Docker and Windows/macOS package builds, that I can understand and maintain instead of relying on something external that you are managing.

2. I don't like the way the snap package is making changes to the source code by changing the `check_github` setting. I would much rather prefer handling restricted/disabled settings the same way it is handled with the Docker container or the Windows/macOS packages. Tautulli is able to identify that is is running in a Docker container or as a bundled package and dynamically disable or hide settings based on that information. In the same way Tautulli should be able to identify it is running as a snap package and dynamically change settings as required.

If you can expand this pull request to include those two items, then I will review again.

I am also not opposed to adding the snap package to our [installation wiki](https://github.com/Tautulli/Tautulli-Wiki/wiki/Installation) as a third party option similar to the other options that are currently listed. In that case we will redirect any users with issues or questions about the snap package to your link for you to provide support.

capruro:
Hi @JonnyWong16, thank you for your feedback.
I'd like to clarify your remarks:

1. I fully understand your concern and I can support you in this automation phase.
Snapcraft support automatic build from GitHub simply linking your account, every time you commit on master branch it will trigger a new build on snapcraft.io that will be automatically published in the store (you can choose if you want to publish directly to stable/rc/beta/edge channel and then manually move it between them) only master branch can be used by snapcraft.io. You can check their documentation https://snapcraft.io/build

If you want, you can create your own account on snapcraft.io and I will ask them to migrate the snap under your account, to make it official.

2. I came already with this problem also on Discord and I agree with you.
Tautulli should be able to identify that is is running in a snap and automatically disable and/ or hide settings based on that information. Mine was a quick workaround to avoid a massive edit of the source that I would have to maintain and fix every release. Since I don't know your code very well and which is the best way to achieve this result I would be glad if you can work on modify tautulli source to have a variable or script that can be triggered during the snapcrafting to have a clean way to hide/disable autoupdate.

Let me know what you think.
Thank you

JonnyWong16:
Please transfer ownership of the Sanp to my account. Same username as GitHub.

capruro:
Hi @JonnyWong16, thank you for your commitment. 
Request created https://forum.snapcraft.io/t/please-transfer-tautulli-snap-upstream/21732

Cheers!

JonnyWong16:
I think this is good to merge now. Thanks.

PR #1378, opened 2020-08-24
Add negative margin to sections with fixed cards

dotsam:
Cards have a 25px right margin, this sets a negative margin on those sections so that cards will sit within the bootstrap row without needing an extra 25px. I believe this is only applicable to the index view, but the might happen somewhere else that I missed.


JonnyWong16:
Thanks.

dotsam:
Hmmm, I seem to have caused some horizontal scroll overflow issues on mobile. I’ll take a look tomorrow and submit a PR to fix.

PR #1381, opened 2020-09-21
Avoid daemon forking with systemd

MichaIng:
systemd units allow to run processes in foreground while daemonization is done on systemd service level when using Type=simple (default). This allows systemd to reliably track the service state, signals and could catch outputs, i.e. it is possible to remove "--quiet" to have Tautulli logging to systemd journal (journalctl) additionally or alternatively to log files.

In case of Type=forking, a PID file is required to allow system reliably determine the service state, which would be an alternative, but has no real advantage. The solution with "GuessMainPID=no" allows systemd to correctly determine the service active state, but e.g. when it is killed, it is seen as "Succeeded." since systemd cannot track the exit code or signal.
__________________
Probably this is all well known and there is some good reason for the current forking solution, but it is better to ask/discuss that based on an actual alternative suggestion 🙂.

### DIFF
--- a/init-scripts/init.systemd
+++ b/init-scripts/init.systemd
@@ -52,9 +52,7 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-ExecStart=/usr/bin/python3 /opt/Tautulli/Tautulli.py --config /opt/Tautulli/config.ini --datadir /opt/Tautulli --quiet --daemon --nolaunch
-GuessMainPID=no
-Type=forking
+ExecStart=/usr/bin/python3 /opt/Tautulli/Tautulli.py --config /opt/Tautulli/config.ini --datadir /opt/Tautulli --quiet --nolaunch
 User=tautulli
 Group=tautulli
 Restart=on-abnormal


JonnyWong16:
I don't understand systemd enough to make a judgment call on this, or know if there are other implications with this change. How does this affect all current installations using the old systemd script?

MichaIng:
> How does this affect all current installations using the old systemd script?

Not at all since the install of the systemd unit is done manually anyway, so changing it upstream has no effect on any existing instance. Also running the old systemd unit of course remains working just fine.

This is only about the way systemd calls the executable so that `systemctl` is able to tell you the exact service state, including exit codes or signals in case. That is more reliable when the process itself does not fork into background and possible since systemd itself forks into background.

What I don't know is if running Tautulli executable without `--daemon` flag has any other relevant implications aside of not forking into background on/after startup. That is the only thing that changes about the finally running process. But I didn't recognise any negative side effects at least, and of course there shouldn't as running it in foreground for testing or debugging or even at regular basis on a dedicated console should work just fine 😉.
__________
But really, this isn't anything important, both versions work just fine. I just see systemd units with `Type=forking` + either PID file or another construct as a result of moreless copying the behaviour of the SysV service script, where this is simply required. Since on systemd it is not, one has some benefit or at least can simplify things a bid.

JonnyWong16:
I think I understand now. I revisited the daemon code in Tautulli and see that the PID file is not even being created when using the current systemd script (`ExecStart=` is missing the `--pidfile` argument). And `GuessMainPID=no` means systemd is not tracking the Tautulli process at all.

So the two options are:
1. Don't fork the process so it can be tracked by systemd. (This is what you have done in this PR.)
    * Remove `--daemon` from `ExecStart=`
    * Remove `GuessMainPID=no`
    * Remove `Type=forking`
2. Create the PID file so systemd can track the process.
    * Add `--pidfile /opt/Tautulli/tautulli.pid` to `ExecStart=`
    * Add `PIDFile=/opt/Tautulli/tautulli.pid`
    * Remove `GuessMainPID=no`
    * Keep `Type=forking`

Is my understanding correct? If so, then I agree that option 1 makes more sense and simplifies things. I will find some time to test these changes.

MichaIng:
Yes exactly. `2.` is basically what is known and required for SysVinit services, probably other service managers as well. `1.` is the expected default on systemd, at least when considering that `Type=simply` is the default service type which expects the main process to run in foreground.

I'm not sure how `GuessMainPID` works, also the exact algorithm is not documented: https://www.freedesktop.org/software/systemd/man/systemd.service.html#GuessMainPID=
At least with Tautulli is comes to the wrong conclusion that the main process has exited after the Python script forks into background, so it obviously guesses wrong 😄. Using `GuessMainPID=no` leads to systemd detecting the service is active, because there are associated/child processes active, which can be even seen when running `systemctl status tautulli`, but since it does not know or guess the main process, it does not track any exit codes or signals (when Tautulli crashes, it simply reports "exited" without any success/failure state) and I think systemctl stop/restart might not work very well because it does not know which process to send the SIGTERM signal to.

JonnyWong16:
I tested this and everything seems to work. Thanks.

PR #1382, opened 2020-10-07
Fix margins on dashboard cards

dotsam:
Another better attempt at #1378. Flipped around the margins on the x-axis this time, also removed the styling of ::-webkit-scrollbar as it is wildly inconsistent between browsers and desktop/mobile. Tested desktop/mobile, safari/chrome, and no overflow issues this time.

### DIFF
--- a/data/interfaces/default/css/tautulli.css
+++ b/data/interfaces/default/css/tautulli.css
@@ -757,9 +757,8 @@ a .users-poster-face:hover {
     float: left;
     position: relative;
     height: 290px;
-    min-width: 350px;
     max-width: 500px;
-    margin-right: 25px;
+    margin-left: 25px;
     margin-bottom: 25px;
 }
 .dashboard-activity-container {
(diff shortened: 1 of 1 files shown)


JonnyWong16:
Thanks. It looks like the margin fix works on desktop and mobile now.

Please don't change the scrollbar. I personally use Chrome for everything and I can't stand the ugly scrollbar. I'm fine with other browsers showing the default scrollbar.

dotsam:
I always forget that scroll bars aren't as simple as they used to be... the default on macOS for a while now has been to hide scrollbars until the use is scrolling, and even then, to have the scroll indicator overlay content, not displace it.

So the CSS to style the scrollbars here actually forces always-visible scrollbars on macOS in defiance of the user preference. In the perfect world, there would be a pseudo class or something to detect this, but there's not, so I'll look in to implementing something like the solution proposed here to detect that and not force the style when scrollbars aren't being shown: https://www.filamentgroup.com/lab/scrollbars/

For now, I'll add back in the scrollbar styling so this PR can be merged and I'll make another one for the scrollbars.

JonnyWong16:
Looks good. Thanks.

PR #1383, opened 2020-10-21
Fix styling of Apple data-detectors in newsletters

zheileman:
The existing style was not properly targetting the links Apple inject when (wrongly, in this case) detecting phone numbers in newsletters.

This has no effect in any other platform or device.
The numbers are still clickable, couldn't fine a way to disable the functionality completely (tried the `format-detection` meta tag with no luck), but at least the styles are not changed anymore.

I tested this on iPhone and iPad and you can see how it looks before and after the change below.

### Before
example1
example2

### After
example1b
example2b

### DIFF
--- a/data/interfaces/newsletters/recently_added.html
+++ b/data/interfaces/newsletters/recently_added.html
@@ -521,7 +521,7 @@
                 line-height: 100%;
             }
 
-            .apple-link a {
+            a[x-apple-data-detectors] {
                 color: inherit !important;
                 font-family: inherit !important;
                 font-size: inherit !important;
--- a/data/interfaces/newsletters/recently_added.internal.html
+++ b/data/interfaces/newsletters/recently_added.internal.html
@@ -521,7 +521,7 @@
                 line-height: 100%;
             }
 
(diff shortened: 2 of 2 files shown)


JonnyWong16:
I think the `.apple-link a` CSS class should be kept so that both will be used (maybe on older iOS devices?)

```
.apple-link a,
a[x-apple-data-detectors] {
    color: inherit !important;
    font-family: inherit !important;
    font-size: inherit !important;
    font-weight: inherit !important;
    line-height: inherit !important;
    text-decoration: none !important;
}
```

zheileman:
@JonnyWong16 The old class was not being used, that is why I removed it. It could never work in first place.

The way the old class is written requires knowing where an Apple device will "detect" a phone number (or address, or email...) and applying the `apple-link` class to their container. For example the containers for the movie year and the movie duration.

However this new class is meant to work by only targeting the links that are created by Apple, no matter where they are.

JonnyWong16:
Oh sorry, my mistake. Yeah, you're right.

PR #1384, opened 2020-11-12
Update reference link to CherryPy auth tool

krowvin:
Link is broken. Added two alternate links to the same information.

### DIFF
--- a/plexpy/webauth.py
+++ b/plexpy/webauth.py
@@ -16,7 +16,6 @@
 #  along with Tautulli.  If not, see <http://www.gnu.org/licenses/>.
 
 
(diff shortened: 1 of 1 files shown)


plexpy/webauth.py
```diff
@@ -16,7 +16,8 @@
 #  along with Tautulli.  If not, see <http://www.gnu.org/licenses/>.
 
 
-# http://tools.cherrypy.org/wiki/AuthenticationAndAccessRestrictions
+# http://web.archive.org/web/20170210040849/http://tools.cherrypy.org/wiki/AuthenticationAndAccessRestrictions
```
Arcanemagus: This is just a reference, there's no need to link to the wayback archive since the Github archive is sufficient.
```suggestion
```

Arcanemagus (changes requested):

PR #1389, opened 2021-03-01
add most active libraries card and total plays to home

herby2212:
## Description

Addition of a home card called "Most Active Libraries" which shows the total plays/duration of a library in the selected time period (days).
The card is sorted by highest amount of plays/duration for the libraries.
The picture changes, when hovering over a library name, according to the library type.

Addition of a total play count (not affected by choosen time period) directly in the Library Statistics of the home page for all library types (excluding photos). As a side-result a easy comparison between, plays for a time period and in total for a library, is now possible on one page.

tautulli_most_active_libraries

## Type of change

## Checklist:

### DIFF
--- a/data/interfaces/default/home_stats.html
+++ b/data/interfaces/default/home_stats.html
@@ -112,7 +112,7 @@ DOCUMENTATION :: END
             % elif stat_id == 'most_concurrent':
             <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon stats-${stat_id} transparent hidden-xs"></div>
             % elif stat_id == 'top_libraries':
-            <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon library-${row0['type']} hidden-xs"></div>
+            <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon library-${row0['section_type']} hidden-xs"></div>
             % endif
             <div class="dashboard-stats-info-container">
                 <div id="stats-title-${stat_id}" class="dashboard-stats-info-title">
@@ -135,7 +135,7 @@ DOCUMENTATION :: END
                             % for row in top_stat['rows']:
                             <li class="dashboard-stats-info-item ${'expanded' if loop.index == 0 else ''}" data-stat_id="${stat_id}"
                                 data-rating_key="${row.get('rating_key')}" data-guid="${row.get('guid')}" data-title="${row.get('title')}"
-                                data-art="${row.get('art')}" data-thumb="${row.get('thumb')}" data-platform="${row.get('platform_name')}" data-library-type="${row.get('type')}"
+                                data-art="${row.get('art')}" data-thumb="${row.get('thumb')}" data-platform="${row.get('platform_name')}" data-library-type="${row.get('section_type')}"
                                 data-user_id="${row.get('user_id')}" data-user="${row.get('user')}" data-friendly_name="${row.get('friendly_name')}" data-user_thumb="${row.get('user_thumb')}"
                                 data-last_watch="${row.get('last_watch')}" data-started="${row.get('started')}" data-live="${row.get('live')}">
                                 <div class="sub-list">${loop.index + 1}</div>
@@ -162,7 +162,10 @@ DOCUMENTATION :: END
                                     % elif stat_id == 'most_concurrent':
                                     ${row['title']}
                                     % elif stat_id == 'top_libraries':
-                                    ${row['library_name']}
+                                    <% library_href = page('library', row['section_id']) %>
+                                    <a href="${library_href}" title="${row['section_name']}">
+                                        ${row['section_name']}
+                                    </a>
                                     % endif
                                 </div>
                                 <div class="sub-count">
--- a/data/interfaces/default/home_stats.html
+++ b/data/interfaces/default/home_stats.html
@@ -76,6 +76,8 @@ DOCUMENTATION :: END
         <div id="stats-background-${stat_id}" class="dashboard-stats-background" style="background-image: url(${page('pms_image_proxy', row0['art'], row0['rating_key'], 500, 280, 40, '282828', 3, fallback=fallback)});">
         % elif stat_id == 'top_platforms':
         <div id="stats-background-${stat_id}" class="dashboard-stats-background platform-${row0['platform_name']}-rgba no-image">
+        % elif stat_id == 'top_libraries':
+        <div id="stats-background-${stat_id}" class="dashboard-stats-background" style="background-image: url(${page('pms_image_proxy', row0['art'], None, 500, 280, 40, '282828', 3, fallback='art')});">
         % else:
         <div id="stats-background-${stat_id}" class="dashboard-stats-background flat">
         % endif
@@ -112,7 +114,11 @@ DOCUMENTATION :: END
             % elif stat_id == 'most_concurrent':
             <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon stats-${stat_id} transparent hidden-xs"></div>
             % elif stat_id == 'top_libraries':
-            <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon library-${row0['section_type']} hidden-xs"></div>
+                % if row0['thumb'].startswith('http'):
+                <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat hidden-xs" style="background-image: url(${page('pms_image_proxy', row0['thumb'], None, 80, 80)});"></div>
+                % else:
+                <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon library-${row0['section_type']} hidden-xs"></div>
+                % endif         
             % endif
             <div class="dashboard-stats-info-container">
                 <div id="stats-title-${stat_id}" class="dashboard-stats-info-title">
--- a/data/interfaces/default/home_stats.html
+++ b/data/interfaces/default/home_stats.html
@@ -25,7 +25,7 @@ grandparent_thumb       Returns location of the item's thumbnail. Use with pms_i
 rating_key              Returns the unique identifier for the media item.
 title                   Returns the title for the associated stat.
 
-== Only if 'stat_id' is 'top_tv' or 'top_movies' or 'top_music' or 'top_user' or 'top_platform' ==
+== Only if 'stat_id' is 'top_tv' or 'top_movies' or 'top_music' or 'top_user' or 'top_platform' or 'top_libraries' ==
 total_plays             Returns the count for the associated stat.
 total_duration          Returns the total duration for the associated stat.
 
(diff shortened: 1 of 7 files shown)


herby2212:
Greetings,
I will remove total plays in the library statistics as it doesn't look good with multiple libraries for the same categorie with the same number under each other and there are already enough numbers at the tv show and music libraries.

So this pull request is only for the Most Active Libraries home card.

JonnyWong16:
That was my first thought as well especially for TV show libraries. `Shows / Seasons / Episodes / Plays` is too much on the card.

herby2212:
Total plays removed. Ready for review.

data/interfaces/default/home_stats.html
```diff
@@ -111,6 +111,8 @@
             <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon platform-${row0['platform_name']} transparent hidden-xs"></div>
             % elif stat_id == 'most_concurrent':
             <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon stats-${stat_id} transparent hidden-xs"></div>
+            % elif stat_id == 'top_libraries':
+            <div id="stats-thumb-${stat_id}" class="dashboard-stats-flat svg-icon library-${row0['type']} hidden-xs"></div>
```
JonnyWong16: Libraries can have a custom thumbnail (Library > Edit) so that should be shown here and fallback to the svg icon if there is no custom thumbnail.

data/interfaces/default/home_stats.html
```diff
@@ -159,6 +161,8 @@ <h4>${top_stat['stat_title']}</h4>
                                     ${row['platform']}
                                     % elif stat_id == 'most_concurrent':
                                     ${row['title']}
+                                    % elif stat_id == 'top_libraries':
+                                    ${row['library_name']}
```
JonnyWong16: This should be a link to the library page similar to the top users.

plexpy/datafactory.py
```diff
@@ -888,6 +890,39 @@ def calc_most_concurrent(title, result):
...
+
+                for item in result:
+                    library_item = library_data.get_watch_time_stats(section_id=item['section_id'], grouping=None , query_days=time_range)
+
+                    row = {
+                        'total_plays': library_item[0]['total_plays'],
+                        'total_duration': library_item[0]['total_time'],
+                        'type': item['section_type'],
+                        'library_name': item['section_name']
+                    }
```
JonnyWong16, line 937: To be consistent with the rest of the API, this should use `section_type` and `section_name` instead of `type` and `library_name`.

The following values should also be included.

```
section_id
last_play
thumb
art
```

See `get_library_stats` below for the logic to get the thumb and art.

I don't remember exactly why I added all the additional blank values (see top users), but I think it was for some API compatibility reason.

plexpy/datafactory.py
```diff
@@ -888,6 +890,39 @@ def calc_most_concurrent(title, result):
                                    'stat_title': 'Most Concurrent Streams',
                                    'rows': most_concurrent})
 
+            elif stat == 'top_libraries':
+                top_libraries = []
+
+                try:
+                    query = 'SELECT section_id, section_name, section_type ' \
+                        'FROM library_sections'
```
JonnyWong16: Deleted libraries should be excluded.

```sql
WHERE deleted_section = 0
```

JonnyWong16 (changes requested):

data/interfaces/default/index.html
```diff
@@ -810,6 +810,10 @@ <h4 class="modal-title">Unable to Update</h4>
                 $('#stats-background-' + stat_id).removeClass(function (index, className) {
                     return (className.match (/(^|\s)platform-\S+/g) || []).join(' ');
                 }).addClass('platform-' + $(elem).data('platform') + '-rgba');
+            } else if (stat_id == 'top_libraries') {
+                $('#stats-thumb-' + stat_id).removeClass(function (index, className) {
+                    return (className.match (/(^|\s)library-\S+/g) || []).join(' ');
+                }).addClass('library-' + $(elem).data('library-type'));
```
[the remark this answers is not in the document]
herby2212: I would stay without the library background image so there is a clean look similar to the `Most Active Users` card.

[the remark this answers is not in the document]
JonnyWong16: Users don't have a background but libraries do. I actually want to add a background colour for users similar to the `Most Active Platforms` but that's a bit more difficult.

plexpy/datafactory.py
herby2212: Good point with the naming convention.
I saw those empty values at different `stat`s, but sure I will add them.

data/interfaces/default/index.html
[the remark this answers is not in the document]
herby2212: I got that, but I personally would go with a cleaner look (without the background image) to answer your question as it seems more fitting for a stats card and not item based (like last watched). I will make the changes for the library background image and see how it looks.
Just as a Idea: Would it be possible to set the background color for a user based on the users icon color with the largest share.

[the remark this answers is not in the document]
JonnyWong16: Yes, but it requires image processing to determine the dominant colour in an image.

data/interfaces/default/home_stats.html
```diff
@@ -133,7 +135,7 @@ <h4>${top_stat['stat_title']}</h4>
                             % for row in top_stat['rows']:
                             <li class="dashboard-stats-info-item ${'expanded' if loop.index == 0 else ''}" data-stat_id="${stat_id}"
                                 data-rating_key="${row.get('rating_key')}" data-guid="${row.get('guid')}" data-title="${row.get('title')}"
-                                data-art="${row.get('art')}" data-thumb="${row.get('thumb')}" data-platform="${row.get('platform_name')}"
+                                data-art="${row.get('art')}" data-thumb="${row.get('thumb')}" data-platform="${row.get('platform_name')}" data-library-type="${row.get('type')}"
```
herby2212: Shall it stay library-type or be re-named to section-type? I would leave it like that.

JonnyWong16: This is fine.

data/interfaces/default/home_stats.html
herby2212: On what client is an option to edit the library icon or the background image (I tested the web app and Plex for Windows)?
As for me the only way to include those would be editing the database from plex itself, even though I'm not sure if Plex would even take this setting into account for the different clients.

JonnyWong16: Edit the library in Tautulli, not Plex.

herby2212: With this change we have inconsistency in the appearance.
Screenshot_1
This is because thumbnail in the library statistics only change when a http link is provided not a local ressource (in this case it switches back to the default icon class of the library type). Is this intended?

herby2212: That is clear, but currently the custom thumbnail for a library doesn't work at all. The background does, but the default value `/:/resources/movie.png` never worked at least for me. I just realized this now while working on it. It seems the issue comes from `if (thumb.startsWith('http')) {` which messes up the query resulting in a not found image.
I am sorry, if I am missing a point here just want to make sure it works as intended when I'm already working on it.
Edit: The query isn't triggered at all for local ressources (no http link) so the icon will also be the default one for the library type.

JonnyWong16: Yes, it's intended. I ignore the `/:/resources/` icons.

plexpy/config.py
```diff
@@ -546,3 +546,14 @@ def _upgrade(self):
...
+        if self.CONFIG_VERSION == 17:
+            home_stats_cards = self.HOME_STATS_CARDS
+            if 'top_users' in home_stats_cards:
+                top_users_index = home_stats_cards.index('top_users')
+                home_stats_cards.insert(top_users_index, 'top_libraries')
+            else:
+                home_stats_cards.add('top_libaries')
+            self.HOME_STATS_CARDS = home_stats_cards
+
+            self.CONFIG_VERSION = 18
```
JonnyWong16: Newline at end of file.

plexpy/datafactory.py
```diff
@@ -888,6 +890,57 @@ def calc_most_concurrent(title, result):
...
+                    result = monitor_db.select(query)
+                except Exception as e:
+                    logger.warn("Tautulli DataFactory :: Unable to execute database query for get_home_stats: top_libraries: %s." % e)
+                    return None
+
+                library_data = libraries.Libraries()
+
+                for item in result:
+                    library_item = library_data.get_watch_time_stats(section_id=item['section_id'], grouping=None , query_days=time_range)
+
```
JonnyWong16, line 915: Don't include libraries with 0 plays/duration.

```python
if library_item[0]['total_plays'] == 0 and library_item[0]['total_time'] == 0:
    continue
```

plexpy/datafactory.py
```diff
@@ -888,6 +890,57 @@ def calc_most_concurrent(title, result):
...
+                        'art': library_art
+                    }
+
+                    top_libraries.append(row)
+
+                home_stats.append({
+                    'stat_id': stat,
+                    'stat_type': sort_type,
+                    'stat_title': 'Most Active Libraries',
+                    'rows': session.mask_session_info(sorted(top_libraries, key=lambda k: k[sort_type], reverse=True), mask_metadata=False)
```
JonnyWong16: Limit to 10 rows, also add a line break for the long line

```python
'rows': session.mask_session_info(
    sorted(top_libraries, key=lambda k: k[sort_type], reverse=True)[:10],
    mask_metadata=False)
```

JonnyWong16 (changes requested):
Looks good. Just a few minor details.

I also fixed the `pms_image_proxy` function for `http` images.

plexpy/datafactory.py
```diff
@@ -888,6 +890,62 @@ def calc_most_concurrent(title, result):
...
+                    }
+
+                    top_libraries.append(row)
+
+                home_stats.append({
+                    'stat_id': stat,
+                    'stat_type': sort_type,
+                    'stat_title': 'Most Active Libraries',
+                    'rows': session.mask_session_info(
+                        sorted(top_libraries, key=lambda k: k[sort_type], reverse=True),
```
JonnyWong16: Missing the limit to maximum 10 rows

```python
sorted(top_libraries, key=lambda k: k[sort_type], reverse=True)[:10]
```

herby2212: Thanks for the notice. Fixed.

JonnyWong16 (approved):

JonnyWong16:
Thanks. Merging this into `nightly` for more testing.

JonnyWong16:
FYI, upgrading failed when the `top_user` card was disabled. Typo in `top_libraries` and lists use the `append()` method, not `add()`. This was my bad for not catching it.

Fix here: e7e5d205a42f05007e63fbff2f1fdd575eb48388

herby2212:
Oh sorry for the trouble caused. You can't catch any error. I see this one on my side for doing such a simple mistake. Thanks for providing the quick fix.

Issue #1416, opened 2021-04-16
Dynamic MQTT Topic


ccutrer:
Right now MQTT notifications are all delivered to a single topic. It'd be easier to process on the other end if you could hook up to individual topics. Either allow putting arbitrary parameters into the topic (i.e. `tautulli/{machine_id}/playback_state`), or use the notification subject as the topic. In either case, you may want an option to pass through the body directly (instead of JSON encoding a body/topic/subject object). This could avoid needing to deal with JSON-packed messages entirely on the other end.

ccutrer:
Really? No comment because I didn’t follow the template that makes zero sense for a feature request?! Would you be interested in a PR at least?

Hellowlol:
It make senses to me. I agree with his decision about closing this issue. You got a automated response, if you can’t be bothered with making a proper issue within 12 days.. Why should he care? We all do this in our spare time, and there are only to much time and attention you can give a task. I’m sure any open source project likes a PR, with the template filled out and to the correct branch.

ccutrer:
I apologize for being flippant. I was both tired and apparently moody this morning. My memory was that the template was things like "describe your operating environment" which is completely unnecessary information for a feature request. After re-inspecting the template, it does ask on-topic questions. Questions which I felt I had addressed in my single paragraph. The bot auto-response felt like an overzealous bot, and feeling like I had addressed it already. GitHub did _not_ send a notification that a human had looked at the issue and added the invalid-template label, so I figured I'd just give it some time for someone to look at it. I definitely understand it's a best-effort volunteer project. I own several of those myself, and sometimes it can take me weeks for months to get back to a question. I also am the primary contributor of a very active project for my job that happens to be open source, and get that GitHub issue trackers Signal-to-Noise can be low. I was just upset that the only notification I got was "issue closed", without even a "please fill out the template completely, including leaving in the template headings". I also went for slightly less formal since the project doesn't seem to be a process nazi, given the lack of a comprehensive test suite (something that happens a lot for my own side projects!).

Anyhow, my need for improved integration with Plex is becoming more necessary, and my other projects before hand (doing the actual real-world labor of finishing my home theater) is wrapping up, so I'll probably work on this myself in the next week or so, and shoot a PR over. Honestly, that was my intent the whole time - just wanted to get a feel for if the feature seemed reasonable before I spent time on it.

JonnyWong16:
[...]

PR #1424, opened 2021-04-29
MQTT improvements

### LINKED ISSUE #1415: JSON Escaping For Notifications
The MQTT notifier is useful for a machine to quickly receive and process notifications. It even already wraps the "message" subject, body, and topic into JSON. However, it's difficult to parse the message body by machine. Ideally support for just listing the parameters you want, and they get stuffed into the root JSON object would be great, but I suspect that would be difficult since the notification system is generic. Instead, how about a text modifier (similar to the upcase and downcase modifiers) that escapes the parameter for insertion into JSON. I.e.:

```json
{
  "username": "{username}",
  "title": {title!json}
}
```

(notice no quotes around the {title} - it would be expected that they would be added, because it's outputting a valid JSON string)

So I'd still have to build up my body and write the JSON, and on the receiving end I'll have to double-decode the JSON (since the body will be encoded as a string in the entire JSON message, but at least it would be machine processable without resorting to any tricks.

ccutrer:
## Description

This is a series of changes to improve the machine readability of notifications sent via MQTT. In particular, being able to send to multiple topics based on any parameter (i.e. I plan to include the {player_id} in the topic so that if I want to do automations for what's happening in a particular room, I only have to subscribe to that topic), and being able to either send bare 
text not wrapped in JSON for simpler handling at the recipient OR being able to build a full JSON object with proper escaping to bundle a lot of information in a single notification.

Fixes #1415 and #1416.

A couple of notes:
 * I'm more than willing to split each commit into its own PR if you would prefer to review that way.
 * I didn't love the deferred import of build_notify_text, but it was necessary to avoid a circular dependency. Possibly break that method out into a separate module? Also, possibly refactor it so that it doesn't need to do repetitive work on both subject and body, and just call it twice.
 * It might make more sense to just have the subject correspond to the MQTT topic, and the body the payload, but that would be a breaking change (there would no longer be a topic configuration setting at all). This would also eliminate the need for the omit_subject option.

## Type of change

## Checklist:

### DIFF
--- a/data/interfaces/default/settings.html
+++ b/data/interfaces/default/settings.html
@@ -1883,6 +1883,25 @@ Rating: {rating}/10   --> Rating: /10
 {Rating: &lt;rating&gt;/10} --> </pre>
                     </div>
                     <div>
+                        <h4>JSON</h4>
+                    </div>
+                    <div style="padding-bottom: 10px;">
+                        <p class="help-block">
+                            Notification parameters with the <span class="inline-pre">!j</span> modifier will be dumped as a JSON string.
+                            Note that if you want to include literal braces in your body (to build a JSON object), they will need to be
+                            doubled to avoid them being interpreted as parameter references.
+                        </p>
+                        <p><strong style="color: #fff;">Example:</strong></p>
+                        <pre>{player}              --> John's "Theater"
+{player!j}            --> "John's \"Theater\"" </pre>
+                        <p><strong style="color: #fff;">Example building full JSON object:</strong></p>
+                        <pre>{{ "title": {title!j},
+ "media_type": {media_type!j} }}       -->
+
+{ "title": "Paw Patrol - Pups Save a School Day",
+  "media_type": "episode" } </pre>
+                    </div>
+                    <div>
                         <h4>Combined</h4>
                     </div>
                     <div>
@@ -1892,7 +1911,7 @@ Rating: {rating}/10   --> Rating: /10
                         <ol class="help-block">
                             <li>Prefix</li>
                             <li>Parameter</li>
-                            <li>Case Modifier</li>
+                            <li>Case Modifier/JSON</li>
                             <li>List Slicing</li>
                             <li>Suffix</li>
                         </ol>
--- a/plexpy/notification_handler.py
+++ b/plexpy/notification_handler.py
@@ -1783,6 +1783,8 @@ class CustomFormatter(Formatter):
             return str(value).lower()
         elif conversion == 'c':  # capitalize
             return str(value).title()
+        elif conversion == 'j': # JSON
+            return json.dumps(str(value))
         else:
             return value
 
--- a/plexpy/notifiers.py
+++ b/plexpy/notifiers.py
@@ -2110,14 +2110,25 @@ class MQTT(Notifier):
                        'keep_alive': 60
                        }
 
-    def agent_notify(self, subject='', body='', action='', **kwargs):
-        if not self.config['topic']:
+    def agent_notify(self, subject='', body='', action='', parameters={}, **kwargs):
+        if plexpy.PYTHON2:
+            from notification_handler import build_notify_text
+        else:
+            from plexpy.notification_handler import build_notify_text
+
+        topic = self.config['topic']
+        if not topic:
             logger.error("Tautulli Notifiers :: MQTT topic not specified.")
             return
 
+        # format the topic just like the body
+        _, topic, _ = build_notify_text(subject='',
+                                        body=topic,
+                                        parameters=parameters)
+
         data = {'subject': subject,
                 'body': body,
-                'topic': self.config['topic']}
+                'topic': topic}
 
         auth = {}
         if self.config['username']:
@@ -2125,7 +2136,7 @@ class MQTT(Notifier):
         if self.config['password']:
             auth['password'] = self.config['password']
 
-        single(self.config['topic'], payload=json.dumps(data), qos=self.config['qos'], retain=bool(self.config['retain']),
+        single(topic, payload=json.dumps(data), qos=self.config['qos'], retain=bool(self.config['retain']),
                hostname=self.config['broker'], port=self.config['port'], client_id=self.config['clientid'],
                keepalive=self.config['keep_alive'], auth=auth or None, protocol=self.config['protocol'])
 
@@ -2174,7 +2185,7 @@ class MQTT(Notifier):
                          {'label': 'Topic',
                           'value': self.config['topic'],
                           'name': 'mqtt_topic',
-                          'description': 'The topic to publish notifications to.',
+                          'description': 'The topic to publish notifications to. Can be dynamic by including parameters just like the message body.',
                           'input_type': 'text'
                           },
                          {'label': 'Quality of Service',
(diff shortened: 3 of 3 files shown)


JonnyWong16:
I don't understand the purpose of the `!j` modifier.

Your example:
```
{{ "title": {title!j}, "media_type": {media_type!j} }}
```
Is the exact same as enclosing parameters in quotes:
```
{{ "title": "{title}", "media_type": "{media_type}" }}
```
`json.dumps(str(value))` is the same as just `str(value)` which is just `"{value}"`.

---

Using `build_notify_text` isn't necessary. The topic doesn't need to filtering `<tags>` and all the custom text modifiers. Just format the topic directly. 

```python
topic = self.config['topic'].format(**kwargs['parameters'])
```

---

`omit_subject` should be `incl_subject` to match the other notification agents. Checkboxes should be "check to enable something" and not "check to disable something". It will need to be checked by default so it doesn't break existing configs.

ccutrer:
`json.dumps(str(value))` is _not_ the same as `str(value)`. It does JSON escaping so that it can be machine parseable. Notice the first example, where the param has a double quote embedded. And in the output, that double quote is properly escaped to form a valid JSON string.
```python
>>> json.dumps("stuff \"\\ and more")
'"stuff \\"\\\\ and more"'
```
Double quotes and backslashes are not an issue in a lot of the fields, but they are for user readable/editable metadata like item titles or player names.

As for the topic and subject options, I've been thinking more and I don't love the direction I'm taking here. I'm thinking instead of a `incl_subject` config, have a `use_subject_as_topic` config, thereby ignoring the topic in the notifier config. It means I don't have to bother with any formatting at all, and I also don't have to have a new option that defaults to "on" in order to be backcompat and non-negative.

JonnyWong16:
Yes, it only makes a difference if a value has a double quote. Otherwise `{param!j}` and `"{param}"` are identical. It's not really an intuitive way to just escape a double quote.

I would much rather have a setting to parse the text as json instead similar to webhooks. No json modifier, no omit/include subject, just one checkbox to enable parsing both the subject and body as json.

---

I don't like the idea of `use_subject_as_topic`. I don't want there to be two possible places to enter a topic which will cause confusion.

ccutrer:
I looked at the webhooks stuff, and I have to admit I don't understand how that works to handle escaping issues. And googling it just led to people complaining that the body they entered couldn't be parsed as JSON, and no response as to how it's supposed to be used. It looks like the webhook notifier takes the body (which has already had the parameters formatted in), and parses it as JSON. then it takes the resulting hash and embeds it in some wrapper json (with a key of either json or data, depending on if a header was set). But if {title} has a a double quote or a slash in it, it would be impossible for body to be properly formed JSON _before_ the JSON parsing, thus throwing an error.

> no omit/include subject

So you're saying a single option to say "parse the subject and body as JSON". I'm not sure how that accomplishes any of my goals here. Maybe I should illustrate how I have it set up right now, and it's working beautifully with my home automation system.

Here is the configuration for my first notifier:
https://www.dropbox.com/s/17ycoyax5qzs1f4/Screen%20Shot%202021-04-29%20at%209.49.50%20AM.png?dl=0
and
https://www.dropbox.com/s/0ayuxg0pby7arkp/Screen%20Shot%202021-04-29%20at%209.50.23%20AM.png?dl=0

and a second:
https://www.dropbox.com/s/vtnlgj0zcoyhy6d/Screen%20Shot%202021-04-29%20at%209.50.41%20AM.png?dl=0
and
https://www.dropbox.com/s/6zgy2mhpsf9sa3a/Screen%20Shot%202021-04-29%20at%209.50.54%20AM.png?dl=0

I also have two more that publish parent_rating_key and grandparent_rating_key. Then in MQTT, the resulting topics look like this (ignore the value of the root tautulli topic; that one is leftover from testing):
https://www.dropbox.com/s/393w3kix6tqdncr/Screen%20Shot%202021-04-29%20at%209.51.42%20AM.png?dl=0

I plan on adding more fields later, but this is enough for now for my code to identify exactly what is playing and take appropriate actions (it adjusts the zoom on the projector depending on the content's aspect ratio).

This make it easy for my HA system to link one item to tautulli/ce2988369206e191-com-plexapp-android/item_key and one item to tautulli/ce2988369206e191-com-plexapp-android/last_action and not have to do any post-processing. I use OpenHAB, which it _can_ extract values from JSON encoded topics, but it's a bit of a pain. I know other home automation systems can work better stuffing a lot of information into a single topic, but again, that will only work correctly if the JSON is properly encoded.

JonnyWong16:
(Your screenshots are all out of order)

I didn't say remove formatting in the topic. Keep that, but just format it directly like I mentioned above.

I'm just saying have a checkbox so you can enter json directly into the subject and body. The webhook notification agent already has all the code to properly escape quotes and everything and you don't need the double curly brackets.

Example, in the webhook body:
```
{
    "title": "{title}",
    "media_type": "{media_type}"
}
```
will automatically be formatted to
```
{
    "title": "A title with \"quotes\"",
    "media_type": "movie"
}
```
compared to your solution which has extra brackets and modifiers
```
{{
    "title": {title!j},
    "media_type": {media_type!j}
}}
```

ccutrer:
Oh, okay, I thought you were referring to no other checkboxes. I can definitely simplify the topic formatting.

And I definitely agree if your description of webhook's formatting is how it works, that would be far easier. I just don't understand how that's possible, but I'll play around with it. It's been a year or 15 since I've used python for more than just a tweak or two.

ccutrer:
Gah, silly dropbox. I think I pasted before they were uploaded. Sorry. Post updated with correct links.

ccutrer:
Ohhhh, I get it now. `build_notify_text` special cases the webhooks notifier, parses the template as JSON, then formats each individual K/V pair. it then dumps it back to a JSON string properly. OK, I'll hook in there, if configured, for MQTT messages

begunfx:
Can we please get this re-opened?

JonnyWong16:
If you are offering to implement the required changes, the please make those changes and submit a new PR.

begunfx:
I am not.  I was just submitting the issue in hopes that someone would address it.  Since you directed me here, I thought that this was an effort to address that issue, that's why I asked if it could get re-opened.

JonnyWong16:
OP abandoned this PR over a year ago, so any changes will require someone to submit a new PR.

ccutrer:
I'm still watching ;), but haven't been motivated to address anything because I've been busy with many other projects. I might be more motivated to work on it now knowing that someone else would appreciate it.

begunfx:
Thanks for chiming in @ccutrer.  I would definitely appreciate it if this was fixed/improved!

ccutrer:
@JonnyWong16:

I've pushed an updated branch, if you could reopen please. I think I've addressed all of your comments:

 * use a simple `topic.format(**parameters)` instead of using `build_notify_text`
 * use `incl_subject`, defaulting to on, instead of `omit_subject`
 * added an option to parse MQTT bodies as JSON, triggering the same logic as the webhook agent uses, rather than having a format specifier and having to double up literal braces.

begunfx:
Thanks, @ccutrer!  @JonnyWong16, looking forward to your action on this PR.

JonnyWong16:
Please submit a new PR based on the latest `nightly` branch.

ccutrer:
I did rebase on top of the latest nightly, but I'll be happy to open a new PR.

PR #1425, opened 2021-05-01
Config for CheryPy Thread Pool Size

psaab:
In some instances, a static thread pool size of 10 is too small.
Provide a way to modify this via the config.all for large servers

## Description

This fixes quite a lot of http accept stalls due to lack of resources.

## Type of change

## Checklist:

### DIFF
--- a/plexpy/config.py
+++ b/plexpy/config.py
@@ -137,6 +137,7 @@ _CONFIG_DEFINITIONS = {
     'HTTP_RATE_LIMIT_ATTEMPTS': (int, 'General', 10),
     'HTTP_RATE_LIMIT_ATTEMPTS_INTERVAL': (int, 'General', 300),
     'HTTP_RATE_LIMIT_LOCKOUT_TIME': (int, 'General', 300),
+    'HTTP_THREAD_POOL': (int, 'General', 10),
     'INTERFACE': (str, 'General', 'default'),
     'IMGUR_CLIENT_ID': (str, 'Monitoring', ''),
     'JOURNAL_MODE': (str, 'Advanced', 'WAL'),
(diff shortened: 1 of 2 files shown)


StarSmasher44:
I am far from related to any of this, but is 10x not a unreasonably high upgrade? Especially for those that may have weak to terrible CPUs. I wouldn't know how far this can eat away at performance but possibly a x5 increase would see the same benefit without the potential to break someone's processor or anything else that runs on there.

JonnyWong16:
100 is excessive. Make it a hidden advanced setting `HTTP_THREAD_POOL` in the config file and leave the default at 10. Then you can change it yourself.

psaab:
Ok I moved it to a config setting

plexpy/config.py
```diff
@@ -182,6 +182,7 @@ def bool_int(value):
     'SYNCHRONOUS_MODE': (str, 'Advanced', 'NORMAL'),
     'THEMOVIEDB_APIKEY': (str, 'General', 'e9a6655bae34bf694a0f3e33338dc28e'),
     'THEMOVIEDB_LOOKUP': (int, 'General', 0),
+    'THREAD_POOL': (int, 'General', 10),
```
JonnyWong16: Rename to `HTTP_THREAD_POOL`.

JonnyWong16 (changes requested):

psaab:
Changed

JonnyWong16:
... but also move it so it is with the other `HTTP_` settings.

psaab:
Whoops, sorry..

psaab:
ok, i think this is correct now

plexpy/config.py
psaab: done

JonnyWong16 (approved):

JonnyWong16:
Thanks.

Issue #1403, opened 2021-03-30
Library Statistics graph/info (For newly Added Movies/Series)


StarSmasher44:
TL;DR VERSION: I want a counter to show me how many series & movies I added over the last X days, you used to have this feature in a not-so-easily readable graph, this would be a better version.
I just want to know how much movies I added/removed this week, mainly added.

**Is your feature request related to a problem? Please describe.**
Not as much as a problem, I believe this used to be a function in Tautulli, where I could bring up a graph to show me how the amount of movies/series/music has risen in whatever amount of time. And while it wasn't great (Too little info), it worked okay for me to judge how much my library had grown over the past week/weeks after adding and removing things.

**Describe the solution you'd like**
I'd like a way to see the X amount of movies/libraries that I have added in X days or weeks, much like the current system displays stats for X amount of days, I just want to see the amount of movies I had before and after adding and removing a bunch of stuff.

**Describe alternatives you've considered**
None, I like tautulli and like to continue using it, this would just give us a little bit more statistical information, and for me to keep track of how much content I add or remove.

**Additional context**
I guess the easiest way would be to just check for the time it was added, and if that is within X days of your filter it will add to the number.

PR #1472, opened 2021-07-21
Library Statistics Graph Tab

herby2212:
## Description

Addition of a new graph tab called "**Library Statistics**" which features currently two new graphs.

The first one called "**Daily addition by media type**" shows the added movies, shows, seasons and episodes on a per day basis. _The time range is set like for the other graphs._

The second one called "**Addition count by media type**" shows the total count of movies, shows, seasons and episodes added in the set time range split in two categories: Movies and TV which includes shows, seasons and episodes.

**Update 01/08/2021:** https://github.com/Tautulli/Tautulli/pull/1472#issuecomment-898540262
Third graph added called "**Addition count by source resolution**" which shows the total count of movies, shows, seasons and episodes added in the set time range by the different resolutions your media have. _(Currently supported: 4k, 1080p, 720p, 576p, 480p, SD)_

**Update 19/08/2021:** https://github.com/Tautulli/Tautulli/pull/1472#issuecomment-902040474
Fourth graph called "**Library Growth**" shows the library growth by the library stats on a per day basis. It is synced with the "**Daily addition by media type**" graph above so the influence of the additions in regards to the overall growth can be seen instantly.

This feature is work in progress and there are the following things planned before its finished:

- [x] ~~When clicking on a day show the corresponding movies/shows/seasons/episodes (depending on which media type clicked) added on that day.~~ - Moving to a future update.
- [x] Music category including artist, album and track count https://github.com/Tautulli/Tautulli/pull/1472#issuecomment-903318421
- [x] Graph improvements (removing not required options and maybe visual improvements)
- [x] Optimize queries
- [x] Data Connection (new table sourced from Plex) https://github.com/Tautulli/Tautulli/pull/1472#issuecomment-902791388
- [x] Moving the episode count in the "Daily addition by media type" graph on a second y-axis with bigger incremental steps (for better scaling)
- [x] Library Growth/Stats Graph
- [x] Check active libraries
- [x] Handle in Tautulli deleted libraries https://github.com/Tautulli/Tautulli/pull/1472/commits/f87fce6ab9321a2f64ecfb1789f16fcfb1699137

I decided to open this request now so improvements, ideas or changes required (that come up through a review for example) can be implemented while in development.

_This feature is based on the ideas mentioned in issue https://github.com/Tautulli/Tautulli/issues/1403._

#### Library Statistics Graph Tab
General Tab that might get expanded with additional graphs related to library statistics.

tautulli library statistics graph updated

#### Addition count by media type
tautulli addition count by media type
_TV category combines Shows, Seasons and Episodes count._

#### Daily addition by media type
tautulli daily addition count by media type 3 second axis
_Updated version with Episode count moved to a second axis._
tautulli daily addition count by media type 2
_Broader view of Movies, Shows and Seasons added over time. Episodes filtered out here for a better overview._

## Type of Change

## Checklist


herby2212:
**Addition count by resolution** [Added 01.08.2021]
addition count by source resolution v2
_Shows the total count of movies, shows, seasons and episodes added in the set time range by the different resolutions your media have. (Currently supported: 4k, 1080p, 720p, 576p, 480p, SD)_

JonnyWong16:
The graphs are a good idea, but we really shouldn't be using the `recently_added` table for the data. That is only a temporary table that can get erased.

A new permanent table should be created to cache the libraries.

herby2212:
> The graphs are a good idea, but we really shouldn't be using the `recently_added` table for the data. That is only a temporary table that can get erased.
> 
> A new permanent table should be created to cache the libraries.

Oh. Under what circumstances does the table get erased? As the table is quite good to analyse the library growth. An alternative would be to create a new permanent table for this and get the initial data from the Plex server itself. Which in the end could also just be done by making a permanent table based on `recently_added`. Though I'm not sure if this table is still in tact for all users.

herby2212:
**Library Growth** graph synced with **Daily addition by media type** graph [Added 19.08.2021]

This fourth graph called "**Library Growth**" shows the library growth by the library stats on a per day basis. It is synced with the "**Daily addition by media type**" graph above so the influence of the additions in regards to the overall growth can be seen instantly.

Enabling/disabling of a axis is also synced as well as the local storage (so both graphs get loaded with the same axis selected) and the Episode axis (including yAxis and label).

**Demo:**
tautulli-library-stats-graph-synced

**If your library is a bit older like this one the more significant growth phase can be seen at a higher day count:**
tautulli library growth graph 500 days
_This graph also includes the year in the date which is helpful when inspecting longer time periods._

JonnyWong16:
> Oh. Under what circumstances does the table get erased? 

Clicking this button
image

Or running the `update_metadata.py` script after moving media/libraries.

The `recently_added` table also only stores data from after Tautulli was installed, but I don't see anything in these graphs that would be limited by Tautulli. It's all data from the Plex server.

A new table should be created and sourced directly from Plex. Or a table might not even be required if the data can be loaded live from Plex, but that will have to be tested for performance to see if it's viable.

herby2212:
> > Oh. Under what circumstances does the table get erased?
> 
> Clicking this button
> image
> 
> Or running the `update_metadata.py` script after moving media/libraries.
> 
> The `recently_added` table also only stores data from after Tautulli was installed, but I don't see anything in these graphs that would be limited by Tautulli. It's all data from the Plex server.
> 
> A new table should be created and sourced directly from Plex. Or a table might not even be required if the data can be loaded live from Plex, but that will have to be tested for performance to see if it's viable.

Thanks for the information.
Agreed. I would go with a table in Tautulli as (besides the performance side) it would be bad if the user can't use the graphs, because there is no connection at this moment to Plex. As the rest of Tautulli can still be used even when there is no connection so we should stay coherent here.
I will add the point data connection to the ToDo's.

herby2212:
**Database Implementation** [Added 20.08.2021]

Added the library_stats_item table with a corresponding refresh function and options for refresh on startup and interval + corresponding implementation in the settings (see picture below):
tautulli library stats data refresh settings
Through this table I was able to optimize my queries.

The initial creation of the table and the update takes quit long (which is why I set the min. hour interval to 6).
Times for my around 16000 items (movies + shows + seasons + episodes + tracks + albums + artists):
- Creation: 16min
- Update: 8min

The best would be to disable the refresh on startup by default and only trigger it once after the update (and have the interval running) with something like a special db upgrade function. Still unsure about that so feedback would be appreciated.

JonnyWong16:
What is the change in database size before and after? How many rows in the table?

herby2212:
> What is the change in database size before and after? How many rows in the table?

16694 rows. Datasize before: 67152 KB after: 97764 KB = ~30 MB

herby2212:
**Music Integration** [Added 22.08.2021]

**Daily addition by media type** and **Library Growth**:
tautulli music additions by day
_All other categories disabled for a better visibility._ 

**Additions by media type**:
tautulli music additions by media type

**Done for visual elements/additions.**

herby2212:
Notice: It can happen that some media items have a timestamp that is in the future or in a time period before the server existed. This issue comes from the way how Plex handels the `date added` when scanning in files during the creation of a new library.

> if you create a fresh library and in the course of this library creation you add a folder full of media files, Plex will read the datetime stamp of the media files and use this instead as ‘date added’. 

Taken from https://forums.plex.tv/t/date-added-wrong/302575/12.

A way to solve this would be to set the `added_at` field for those media items in the `library_stats_items` table to the library section `created_at` timestamp upon initial creation of the table.
This field is currently not in the `library_sections` table of Tautulli present, only in Plex.

**UPDATE:** Fixed in commit https://github.com/herby2212/Tautulli/commit/2386cab8b2145a178d473eab40d5b8a908861a74.
Media items with a timestamp older then their corresponding library instead receive the library creation timestamp.

plexpy/graphs.py
```diff
@@ -421,6 +421,407 @@ def get_total_plays_per_hourofday(self, time_range='30', y_axis='plays', user_id
...
+                            '0 AS date_added, ' \
+                            'SUM(CASE WHEN media_type = "movie" THEN 1 ELSE 0 END) AS movie_count, ' \
+                            'SUM(CASE WHEN media_type = "show" THEN 1 ELSE 0 END) AS tv_count, ' \
+                            'SUM(CASE WHEN media_type = "season" THEN 1 ELSE 0 END) AS season_count, ' \
+                            'SUM(CASE WHEN media_type = "episode" THEN 1 ELSE 0 END) AS episode_count, ' \
+                            'SUM(CASE WHEN media_type = "artist" THEN 1 ELSE 0 END) AS artist_count, ' \
+                            'SUM(CASE WHEN media_type = "album" THEN 1 ELSE 0 END) AS album_count, ' \
+                            'SUM(CASE WHEN media_type = "track" THEN 1 ELSE 0 END) AS track_count ' \
+                        'FROM library_stats_items ' \
+                        'WHERE added_at < %s ' \
```
herby2212, line 434: Missing check if section is not deleted in Tautulli `deleted_section = 0`. Implementation open for multiple spots.

herby2212:
Any chance this gets a review in the future? The commits were made in smaller parts and the newly added features got explained with the commits so that is it's easier for you to track the changes made. As I saw that this was a issue with larger feature commits in the past.

PR #1474, opened 2021-07-24
Adding simple retry around Cloudinary upload

Mavarice:
## Description

Adding a simple retry loop around the cloudinary upload
If the try succeeds the else will catch and break the loop, thus ending the retries.
If the try has an exception it will raise a warning, unless the number of retries is at the iteration counter-1 (last attempt) in which case it will raise the exception and send out an error, same as it does today.

## Type of Change

## Checklist

### DIFF
--- a/plexpy/helpers.py
+++ b/plexpy/helpers.py
@@ -865,15 +865,28 @@ def upload_to_cloudinary(img_data, img_title='', rating_key='', fallback=''):
     else:
         _img_title = img_title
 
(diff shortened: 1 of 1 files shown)


JonnyWong16:
Why is Cloudinary failing in the first place? Can we figure that out instead of just trying 3 times as quickly as possible? There's no guarantee that the retries will succeed if there is a larger issue at play.

Issue #1490, opened 2021-08-18
Episode Number Under Recently Watched


Jaybonaut:
**Is your feature request related to a problem? Please describe.**
Inability to see the episode number under 'Recently Watched' on the home page - you can only see the episode name.

**Describe the solution you'd like**
It would be nice to see the season and episode number for tv shows under 'Recently Watched' on the home page.

**Describe alternatives you've considered**
I haven't considered any alternatives, as I am unaware of any that exist.

**Additional context**
None needed.

Jaybonaut:
Sure thing.  Where is the template?

samwiseg0 ?

TheMeanCanEHdian:
When you submit an issue there is a template provided. You can see an example for the feature request [here](https://github.com/Tautulli/Tautulli/issues/new?assignees=&labels=status%3Aawaiting-triage%2C+type%3Aenhancement&template=feature_request.md&title=).

Jaybonaut:
Thanks, done.

herby2212:
@Jaybonaut Something like this?
![tautulli recently watched episode number](https://user-images.githubusercontent.com/12448284/130227324-92c8d5dc-b60f-4019-882f-d11a048c414f.png)

Jaybonaut:
Sure

Issue #1487, opened 2021-08-13
Date-Based Television Shows


martinr92:
**Describe the Bug**
Plex supports a naming schema called "Date-Based Television Shows".  The major difference is that the episodes of a show have no episode numbers but a airing date instead. 
More details can be found here:
https://support.plex.tv/articles/naming-and-organizing-your-tv-show-files/

**Steps to Reproduce**
Tautulli is not handling / showing this correctly. E.g. here is no episode number available:
<img width="526" alt="2021-08-13_18-43-55" src="https://user-images.githubusercontent.com/658036/129392824-b1f795e0-aee0-4bf6-aaff-c9767b920263.png">
<img width="345" alt="2021-08-13_18-54-10" src="https://user-images.githubusercontent.com/658036/129393443-e632ed3a-4dc4-4ab0-a415-7415443ca90f.png">

**Expected Behavior**
It should be handled like in the Plex WebUI. Here it shows the date:
<img width="333" alt="2021-08-13_18-43-45" src="https://user-images.githubusercontent.com/658036/129392969-e05b2354-1dff-46f2-be64-7a9d6799d7b3.png">

**Relevant Settings**
A plex library of type *Series* with the episode naming schema liked above:
```
/TV Shows/ShowName/Season 02/ShowName – 2011-11-15 – Optional_Info.ext
```

**Tautulli and System Info (see Tautulli settings page)**
- Version: 2.7.5
- Git Branch: master
- Git Commit Hash: 5fbe9d476b8dec441869085a195794d9efb1ef72
- Platform and Version: Linux 4.19.0-9-amd64 Debian GNU/Linux 10 buster
- Python Version: 3.7.3

martinr92:
Thanks to @herby2212 for working on this.

PR #1492, opened 2021-08-21
Last Watched Identifier

herby2212:
## Description

Addition of an option to have the season and episode number listed for episodes in the Last Watched card on the home page.
This feature can be turned on/off in the settings. (enabled by default)

This feature is based on the issue https://github.com/Tautulli/Tautulli/issues/1490.

### Screenshot

tautulli recently watched episode number
tautulli last watched identifier

## Type of Change

## Checklist


JonnyWong16:
I'm not sure I like the look of the `S00E00` on the card there. And I don't think a new setting is necessary.

Also, this should be done in the UI (mako template) and not modifying the server side (python).

herby2212:
> I'm not sure I like the look of the `S00E00` on the card there. And I don't think a new setting is necessary.
> 
> Also, this should be done in the UI (mako template) and not modifying the server side (python).

I implemented the setting so people can choose if they just want the longer clean name of the episode (like it is currently) or more informations by enabling this setting. It could be disabled by default if this is a pain point for you so only users actively seeking this function can activate it.
With this in mind I think it fits quite well at this point in the `settings` and it is better to give the user the option.

Thanks for the notice regarding the mako template. I looked at the documentation and will push a [commit ](https://github.com/Tautulli/Tautulli/pull/1492/commits/b0338bbd37a28f32f1d3e31b7fc6ab3f2d238fe1) which transfers all logic to the templace. It would be great if you could have a look over it.

JonnyWong16:
Maybe because it's not consistent between movies and TV shows. There is a blank in that spot when it's a movie. What if it is filled in with the movie year? But then it's still not uniform across all rows.

There's also considerations for TV shows with custom season titles (f92ba453c8f1115f9cf401fb2a636cd9ff71e44b) or date-based TV shows (#1487) that don't use the `S00E00` format (although nothing else in the UI accounts for date-based shows yet).

JonnyWong16 (changes requested):

herby2212:
> Maybe because it's not consistent between movies and TV shows. There is a blank in that spot when it's a movie. What if it is filled in with the movie year? But then it's still not uniform across all rows.
> 
> There's also considerations for TV shows with custom season titles ([f92ba45](https://github.com/Tautulli/Tautulli/commit/f92ba453c8f1115f9cf401fb2a636cd9ff71e44b)) or date-based TV shows (#1487) that don't use the `S00E00` format (although nothing else in the UI accounts for date-based shows yet).

If a user activates this option to get more details about the series item I wouldn't add informations (like the year at a movie) just to fill the space. It may look inconsistent, but cleaner in my opinion.

I just tested the case with custom season titles and it will just show the index (which should give the user still a rough estimate about which season the media item is in) and I believe it should stay that way for now. A custom season title would be to long for this value field.

If date-based TV shows are coming `this` along other points would probably need to be re-visited. Currently I think a date-based identifier would still fit with a syntax something like `S2021·11.8.21`, but I would leave this topic for now regarding this PR. 
Though I would be open to support you at this topic if you want.

I cleaned up the spots pointed out and will re-request a review.

herby2212:
@JonnyWong16 I looked at the issue with the missing episode number mentioned in https://github.com/Tautulli/Tautulli/issues/1487. 
The `index` field in the xml of date based tv show elements is missing. I was thinking about the `originallyAvailableAt` field as alternative episode identifier. As this should always match the episode name/identifier.
From a UI perspective it could be checked if there is no index present and the library uses the new Plex TV Agent (as the older ones doesn't support this feature IIRC). The formatting could be done in a similar fashion to Plex.
I hope this helps as I don't know what the current status on this topic is from your site. 

A picture for reference:
date based tv show sessions xml

JonnyWong16:
Yeah, I already know that. It's just _a lot_ of places in the UI that need to change.

herby2212:
> Yeah, I already know that. It's just _a lot_ of places in the UI that need to change.

If you agree on the `originallyAvailableAt` I would open up a PR and start working on this.

JonnyWong16:
Yes, `originallyAvailableAt` is correct when episodes don't have an `index`. Please use ISO date format `YYYY-MM-DD` because Tautulli doesn't have date localization.

herby2212:
Marked as draft until Pull Request https://github.com/Tautulli/Tautulli/pull/1503 is done and merged.

PR #1503, opened 2021-09-07
Date Based TV Show Support

herby2212:
### Continued with PR https://github.com/Tautulli/Tautulli/pull/1960. This PR is outdated and marked as draft.

This PR aims to fix issue https://github.com/Tautulli/Tautulli/issues/1487 and provide a place where we can work & discuss on this topic.

**UI fields needing a change/fixed:**

- [x] Recently Added
- [x] Current Activity
- [x] Library Recently Added
- [x] User Recently Watched
- [x] TV Show History
- [x] Library Media Info
- [x] Users Table
- [x] Info page breadcrumb / header
- [ ] Newsletter templates
- [x] Notification parameters

Please report any occurrences of this issue that aren't listed above.

### Screenshot

tautulli activity date based tv show

## Type of Change

## Checklist

### DIFF
--- a/plexpy/common.py
+++ b/plexpy/common.py
@@ -502,6 +502,7 @@ NOTIFICATION_PARAMETERS = [
              {'name': 'Season Number', 'type': 'int', 'value': 'season_num', 'description': 'The season number.', 'example': 'e.g. 1, or 1-3'},
              {'name': 'Season Number 00', 'type': 'int', 'value': 'season_num00', 'description': 'The two digit season number.', 'example': 'e.g. 01, or 01-03'},
              {'name': 'Episode Number', 'type': 'int', 'value': 'episode_num', 'description': 'The episode number.', 'example': 'e.g. 6, or 6-10'},
+             {'name': 'Episode Number', 'type': 'str', 'value': 'episode_date', 'description': 'The episode number of date based tv shows (in date format).'},
              {'name': 'Episode Number 00', 'type': 'int', 'value': 'episode_num00', 'description': 'The two digit episode number.', 'example': 'e.g. 06, or 06-10'},
              {'name': 'Track Number', 'type': 'int', 'value': 'track_num', 'description': 'The track number.', 'example': 'e.g. 4, or 4-10'},
              {'name': 'Track Number 00', 'type': 'int', 'value': 'track_num00', 'description': 'The two digit track number.', 'example': 'e.g. 04, or 04-10'},
--- a/data/interfaces/default/current_activity_instance.html
+++ b/data/interfaces/default/current_activity_instance.html
@@ -18,6 +18,7 @@ media_type                  Returns the type of session. Either 'track', 'episod
 thumb                       Returns the location of the item's thumbnail. Use with pms_image_proxy.
 bif_thumb                   Returns the location of the item's bif thumbnail. Use with pms_image_proxy.
 art                         Returns the location of the item's artwork
+originally_available_at     Returns the availability date of the media item.  
 progress_percent            Returns the current progress of the item. 0 to 100.
 user                        Returns the name of the user owning the session.
 user_id                     Returns the Plex user id if available.
@@ -62,7 +63,7 @@ DOCUMENTATION :: END
 % if session is not None:
 <%
     from collections import defaultdict
-    from plexpy.helpers import cast_to_int, page, short_season
+    from plexpy.helpers import cast_to_int, page, short_season, format_date_based_show
     from plexpy.common import VIDEO_RESOLUTION_OVERRIDES, AUDIO_CODEC_OVERRIDES, EXTRA_TYPES
     import plexpy
 %>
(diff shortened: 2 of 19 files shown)


JonnyWong16:
You don't need all the `format_date_based_show` functions. `originallyAvailableAt` is already in ISO format `YYYY-MM-DD`. Don't use the dots.

herby2212:
> You don't need all the `format_date_based_show` functions. `originallyAvailableAt` is already in ISO format `YYYY-MM-DD`. Don't use the dots.

I know it is already in the ISO Format. The idea was to keep it in line with the current style (seasons have the dots `S3 · E1` and Plex also for the date based episodes).

herby2212:
@JonnyWong16 
Could you handle the implementation for the Newsletter Templates regarding the date based tv show support? As I'm not familiar with the way they function or the files affected.

So we can complete/progress this Pull Request.

I first tried to contact you via a Discord PM as I would have liked to not clutter up the PR with a message like this.

JonnyWong16:
The date format you see in Plex is based on locale. It will use a dot, dash, or a slash and a different order of year, month, and day depending on where you set your location.

In Tautulli, since there is no localization, the date could be

1. ISO format `YYYY-MM-DD`.
    * Pro: international standard
    * Con: not everyone likes it
2. Follow the "Date Format" in the settings.
    * Pro: user chooses the format they like
    * Con: user wants a different date format for general UI vs. episodes
3. Add a new "Date-Based Episode Format".
    * Pro: user chooses the format they like
    * Con: another setting could be confusing

For Discord, you can just use the `#general` channel.

herby2212:
We should either go with 1. or 3. as 2. would be nothing complete (users can choose but not on a sufficent level). 

So from my perspective:
1 ISO Format:
- Pro: Would be clean and lightweight without any formatting
- Con: Unable to be customized by a user (could be a irregularity for the user if the normal date is customized & customizable but this not)

3 Add a new "Date-Based Episode Format"
- Pro: User can customize it to fit its region and/or match the configured regular date (with a slight difference for example just like by default in Plex)
- Pro: Would be a full feature implementation (incl. setting) instead of a "basic support" implementation (to make it work) as a request for this setting wouldn't seem unlikely once released

I tend to 3. to give users one more option from the start, but I would like to know your opinion.

`For Discord, you can just use the #general channel.`
Okay, wasn't sure if deeper development/PR topics are allowed there _(as there is also no development channel for example)_.

herby2212:
How do we want to display the episode range in the newsletter with date based tv shows?

For some items the listing could work like `2021·08·22, 2021·08·23`.
If there are more episodes take the earliest and latest episode (based on the timestamp) like this `2021·08·22-2021·09·01` maybe.

tautulli newsletter episode range

JonnyWong16:
I need to think about the newsletter.

Let's go with option 3, a separate setting for "Date-Based Episode Format".

herby2212:
> I need to think about the newsletter.
> 
> Let's go with option 3, a separate setting for "Date-Based Episode Format".

Any decision/news about the newsletter?

JonnyWong16:
I'm thinking just the range of dates for the newsletter.

`Season 2, Episodes 2021-01-01 - 2021-06-30`

herby2212:
> I'm thinking just the range of dates for the newsletter.
> 
> `Season 2, Episodes 2021-01-01 - 2021-06-30`

Sounds good and seems like the cleanest solution.

plexpy/common.py
```diff
@@ -502,6 +502,7 @@
              {'name': 'Season Number', 'type': 'int', 'value': 'season_num', 'description': 'The season number.', 'example': 'e.g. 1, or 1-3'},
              {'name': 'Season Number 00', 'type': 'int', 'value': 'season_num00', 'description': 'The two digit season number.', 'example': 'e.g. 01, or 01-03'},
              {'name': 'Episode Number', 'type': 'int', 'value': 'episode_num', 'description': 'The episode number.', 'example': 'e.g. 6, or 6-10'},
+             {'name': 'Episode Number', 'type': 'str', 'value': 'episode_date', 'description': 'The episode number of date based tv shows (in date format).'},
```
herby2212, line 505: Provide a additional notifier variable `episode_date` that is only used for date based tv shows or extend the description of `episode_num` that in case of a date based tv show shows the air date?

JonnyWong16: There's already an `air_date` parameter.

herby2212: Okay, but in case of a item that uses the `date based` system `episode_num` and `episode_num00` will be replaced with `originally_available_at`. Should we include this in the notifier description? That was the idea behind this question/change.

JonnyWong16: No, don't change the behaviour of `episode_num` (in this PR).

herby2212: Okay I will work on that.

herby2212: Just gave it a thought, how would you propose to implement the notifications for date based tv show?

If the user uses `episode_num` and hits a _on date based_ episode the variable then should use the `originally_available_at` as its currently implemented from a logic standpoint in my opinion.

herby2212:
Continued with PR #1960. This PR is outdated and marked as draft.

Issue #1475, opened 2021-07-26
HDR Format listed in Media Info


rasorjb:
**Is your feature request related to a problem? Please describe.**
It is not a problem with the functionality of Tautulli. I am currently having an issue with some movies that were not encoded correctly for HDR. As I have thousands of movies, finding them all using MediaInfo, is basically impossible.

**Describe the solution you'd like**
Add a column to the Media Info window that displays HDR format (HDR10, HDR10+, Dolby Vision, etc)

**Describe alternatives you've considered**
I am currently using mediainfo, just takes forever.

**Additional context**

kuryak1n:
Have you looked at exporting the metadata for a given library?
There are columns for HDR, video codec, color space, bit depth, and several with "DOVI."
You can pull the CSV file into a spreadsheet for easy filtering, sorting, etc.
Not as elegant as your request, but better than searching movie by movie.

PR #1654, opened 2022-02-14
Video Profile (e.g. HDR) listed in Media Info

herby2212:
## Description

This feature adds the video profile as a column to the media info of a library.
_The column can be selected/deselected as usual (see screenshot)._

_This feature is based on the idea mentioned in issue #1475 ._

### Screenshot
tautulli-hdr-profile-media_info_example
_Example of two 4K movies with the `video profile` listed between the `video resolution` and the `video framerate`._

tautulli-hdr-profile-media_info tautulli-hdr-profile-media_info_select_column
_Broader example of different `video profiles` and the option to select/deselect the column._

## Type of Change

## Checklist

### DIFF
--- a/plexpy/pmsconnect.py
+++ b/plexpy/pmsconnect.py
@@ -2724,15 +2724,26 @@ class PmsConnect(object):
                             media_metadata = self.get_metadata(
                                 str(helpers.get_xml_attr(item, 'ratingKey')), output_format='xml')
                             _stream = media_metadata.getElementsByTagName('Stream')[0]
-                            if _stream.hasAttribute('DOVIProfile'):
-                                _videoProfile = 'main 10 HDR DV'
-                            elif _stream.hasAttribute('chromaLocation') and (_stream.getAttribute('colorTrc') == 'bt2020-10' 
-                                    or _stream.getAttribute('colorTrc') == 'arib-std-b67'):
-                                _videoProfile = 'main 10 HDR HLG'
-                            elif helpers.cast_to_int(_stream.getAttribute('bitDepth')) > 8 and _stream.getAttribute('colorSpace') == 'bt2020nc':
-                                _videoProfile = 'main 10 HDR'
+                            _plexVersion = helpers.cast_to_int(plexpy.CONFIG.PMS_VERSION.split('-')[0].translate(str.maketrans('', '', '.')))
+                            #HDR details got introduced with PMS version 1.25.6.5545
+                            if _plexVersion >= 12565545:
+                                displayTitle = str(_stream.getAttribute('extendedDisplayTitle')).lower()
+                                if 'dolby vision' in displayTitle:
+                                    _videoProfile = 'main 10 DV'
+                                elif 'hlg' in displayTitle:
+                                    _videoProfile = 'main 10 HLG'
+                                elif 'hdr' in displayTitle:
+                                    _videoProfile = 'main 10 HDR'
+                                else:
+                                    _videoProfile = 'main 10 SDR'
                             else:
-                                _videoProfile = 'main 10 SDR'
+                                if _stream.hasAttribute('DOVIProfile'):
+                                    _videoProfile = 'main 10 HDR DV'
+                                elif helpers.cast_to_int(_stream.getAttribute('bitDepth')) > 8 and _stream.getAttribute('colorSpace') == 'bt2020nc':
+                                    #Exact HDR version needs PMS version 1.25.6.5545 or higher
+                                    _videoProfile = 'main 10 HDR*'
+                                else:
+                                    _videoProfile = 'main 10 SDR'
 
                         media_info = {'container': helpers.get_xml_attr(media, 'container'),
                                       'bitrate': helpers.get_xml_attr(media, 'bitrate'),
--- a/plexpy/pmsconnect.py
+++ b/plexpy/pmsconnect.py
@@ -2721,7 +2721,7 @@ class PmsConnect(object):
                     for media in item_media:
                         _videoProfile = helpers.get_xml_attr(media, 'videoProfile')
                         if _videoProfile == 'main 10':
-                            media_metadata = PmsConnect().get_metadata(
+                            media_metadata = self.get_metadata(
                                 str(helpers.get_xml_attr(item, 'ratingKey')), output_format='xml')
                             _stream = media_metadata.getElementsByTagName('Stream')[0]
                             if _stream.hasAttribute('DOVIProfile'):
--- a/plexpy/pmsconnect.py
+++ b/plexpy/pmsconnect.py
@@ -2723,6 +2723,7 @@ class PmsConnect(object):
                                       'bitrate': helpers.get_xml_attr(media, 'bitrate'),
                                       'video_codec': helpers.get_xml_attr(media, 'videoCodec'),
                                       'video_resolution': helpers.get_xml_attr(media, 'videoResolution').lower(),
+                                      'video_profile': helpers.get_xml_attr(media, 'videoProfile'),
                                       'video_framerate': helpers.get_xml_attr(media, 'videoFrameRate'),
                                       'audio_codec': helpers.get_xml_attr(media, 'audioCodec'),
                                       'audio_channels': helpers.get_xml_attr(media, 'audioChannels'),
--- a/plexpy/pmsconnect.py
+++ b/plexpy/pmsconnect.py
@@ -2719,11 +2719,21 @@ class PmsConnect(object):
                 if get_media_info:
                     item_media = item.getElementsByTagName('Media')
                     for media in item_media:
+                        _videoProfile = helpers.get_xml_attr(media, 'videoProfile')
+                        if _videoProfile == 'main 10':
+                            media_metadata = PmsConnect().get_metadata(
+                                str(helpers.get_xml_attr(item, 'ratingKey')), output_format='xml')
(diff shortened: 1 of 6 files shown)


JonnyWong16:
The `videoProfile` does not necessarily indicate HDR or Dolby Vison so it is not exactly what #1475 is asking for.

For example, I have a test Dolby Vison file and the `videoProfile` is just "Main 10".

image

Same with HDR, it is just "Main 10" (HDR isn't explicitly anywhere in the XML data).

image

herby2212:
> The `videoProfile` does not necessarily indicate HDR or Dolby Vison so it is not exactly what #1475 is asking for.
> 
> For example, I have a test Dolby Vison file and the `videoProfile` is just "Main 10".
> 
> image
> 
> Same with HDR, it is just "Main 10" (HDR isn't explicitly anywhere in the XML data).
> 
> image

Good catch. I will need to use the metadata this way I can check for DOVI and the HDR version.

herby2212:
The HDR format is now distincly shown which allows to sort by DV/HDR. (https://github.com/Tautulli/Tautulli/pull/1654/commits/db94e0e0ee3cb3eab4949b5001a6f1d072dbb9f9)

tautulli-hdr-profile-media_info_exampleV2

Although I was not able to get a difference between a HDR and HDR 10 Plus movie by the metadata XML.

JonnyWong16:
What about "Main 10" videos that _are not_ HDR?

Example [Jellyfish Video Bitrate Test Files](https://jell.yfish.us/).

---

This is how I determine HDR for the activity card.

https://github.com/Tautulli/Tautulli/blob/master/plexpy/pmsconnect.py#L2149-L2150

However that method does not match Plex exactly. For example, this [test file](https://4kmedia.org/lg-cymatic-jazz-hdr-hlg-uhd-4k-demo/) is not listed as HDR in Plex, but my logic says it is.

image

image

---

The only way I can think of to match Plex exactly is to do a filter for HDR content in Plex and check if the item is in the results.

image

But this is an extra API call. Also, this method won't work for Dolby Vison content.

herby2212:
I don't think that Plex supports HLG at the moment which your test file has. For this file there are two distinct fields `chromaLocation="topLeft"` and `colorTrc="arib-std-b67"`. When muxed into a mkv it turns into `colorTrc="bt2020-10"`.

Plex won't show HDR in both cases.

JonnyWong16:
I updated to Plex Media Server 1.25.6.5545 and it can show "Dolby Vison" and "HLG" now.

image

image

---

Maybe another way to do it is to check if `"HDR" in displayTitle` or `"Dolby Vison" in displayTitle`, etc.

herby2212:
> I updated to Plex Media Server 1.25.6.5545 and it can show "Dolby Vison" and "HLG" now.
> 
> image
> 
> image
> 
> Maybe another way to do it is to check if `"HDR" in displayTitle` or `"Dolby Vison" in displayTitle`, etc.

Finally, took long enough. Yeah this way is easier and we can rely on the Plex logic for the HDR identification.

plexpy/pmsconnect.py
```diff
@@ -2724,15 +2724,26 @@ def get_library_children_details(self, section_id='', section_type='', list_type
...
+                            if _plexVersion >= 12565545:
+                                displayTitle = str(_stream.getAttribute('extendedDisplayTitle')).lower()
+                                if 'dolby vision' in displayTitle:
+                                    _videoProfile = 'main 10 DV'
+                                elif 'hlg' in displayTitle:
+                                    _videoProfile = 'main 10 HLG'
+                                elif 'hdr' in displayTitle:
+                                    _videoProfile = 'main 10 HDR'
+                                else:
+                                    _videoProfile = 'main 10 SDR'
```
herby2212: Identifier for `HDR 10 Plus` open (if supported).

plexpy/pmsconnect.py
```diff
@@ -2724,15 +2724,26 @@ def get_library_children_details(self, section_id='', section_type='', list_type
...
                             else:
-                                _videoProfile = 'main 10 SDR'
+                                if _stream.hasAttribute('DOVIProfile'):
+                                    _videoProfile = 'main 10 HDR DV'
```
herby2212: `main 10 HDR DV` needs to be changed to `main 10 DV` to match the new identification.

plexpy/pmsconnect.py
JonnyWong16: Use `HDR10` instead of `HDR`.

Doesn't look like Plex differentiates HDR10+.

image

[HDR10+ Int'l Space Station Sample](https://kodi.wiki/view/Samples)

plexpy/pmsconnect.py
JonnyWong16: Go with the full name `Dolby Vision` instead of `DV`.

Note it can also be multiple values `Dolby Vision/HLG`, `Dolby Vision/HDR10`.

image

image


[Dolby Vision w/embedded HLG HEVC 60fps](https://kodi.wiki/view/Samples)

JonnyWong16:
I don't think we really care about the video profile, just the type of HDR used. Maybe rename the column to "Video Dynamic Range" or "Dynamic Range"?

plexpy/pmsconnect.py
```diff
@@ -2719,10 +2719,37 @@ def get_library_children_details(self, section_id='', section_type='', list_type
                 if get_media_info:
                     item_media = item.getElementsByTagName('Media')
                     for media in item_media:
+                        _videoProfile = helpers.get_xml_attr(media, 'videoProfile')
+                        if _videoProfile == 'main 10':
+                            media_metadata = self.get_metadata(
+                                str(helpers.get_xml_attr(item, 'ratingKey')), output_format='xml')
+                            _stream = media_metadata.getElementsByTagName('Stream')[0]
+                            _plexVersion = helpers.cast_to_int(plexpy.CONFIG.PMS_VERSION.split('-')[0].translate(str.maketrans('', '', '.')))
```
JonnyWong16: Update `helpers.version_to_tuple`

```py
def version_to_tuple(version):
    return tuple(cast_to_int(v) for v in version.strip('v').replace('-', '.').split('.'))
```

Then just check
```py
if helpers.version_to_tuple(plexpy.CONFIG.PMS_VERSION) >= helpers.version_to_tuple('1.25.6.5545'):
```

plexpy/pmsconnect.py
```diff
@@ -2719,10 +2719,37 @@ def get_library_children_details(self, section_id='', section_type='', list_type
                 if get_media_info:
                     item_media = item.getElementsByTagName('Media')
                     for media in item_media:
+                        _videoProfile = helpers.get_xml_attr(media, 'videoProfile')
+                        if _videoProfile == 'main 10':
+                            media_metadata = self.get_metadata(
+                                str(helpers.get_xml_attr(item, 'ratingKey')), output_format='xml')
+                            _stream = media_metadata.getElementsByTagName('Stream')[0]
```
JonnyWong16, line 2726: Don't need to check video profile.

~~Don't need the extra API call.~~

```py
stream = next(
    (
        helpers.get_xml_attr(stream, 'streamType') == '1'
        for stream in part.getElementsByTagName('Stream')
        for part in media.getElementsByTagName('Part')
    ),
    None
)
```

Actually, this is a big problem. `<Stream>` info isn't returned in the `/library/sections/<sectin_id>/all` endpoint. And doing an additional API call for each item is very expensive. I don't like this at all.

JonnyWong16 (changes requested):

herby2212:
> I don't think we really care about the video profile, just the type of HDR used. Maybe rename the column to "Video Dynamic Range" or "Dynamic Range"?

I would go with `Video Dynamic Range` to keep the naming consistent.

plexpy/pmsconnect.py
herby2212: I know this is why I went with `get_metadata` and added the filter on the video profile to only make a call when needed.

A filter on the `videoCodec` of `<Media>` would also be a solution as dynamic range properties require `hevc` as far as I know. This way the video profile could be removed completely. _-> Done_

It would be good if the `<Stream>` could be requested from the `/library/sections/<sectin_id>/all` endpoint through an option/parameter in the request.

herby2212:
@JonnyWong16 Any news on this? I requested a review some time ago.

JonnyWong16:
From my comment here, I really don't like that this requires an API call per movie. It's just too expensive. 

https://github.com/Tautulli/Tautulli/pull/1654#discussion_r808284160

Until we can figure out a more efficient method, I can't merge this.

herby2212:
> From my comment here, I really don't like that this requires an API call per movie. It's just too expensive.
> 
> [#1654 (comment)](https://github.com/Tautulli/Tautulli/pull/1654#discussion_r808284160)
> 
> Until we can figure out a more efficient method, I can't merge this.

Also if you factor in the filter added in commit https://github.com/Tautulli/Tautulli/pull/1654/commits/72bfb2aa8431343c0c471c01bae3649a085027a1 ([pmsconnect.py](https://github.com/Tautulli/Tautulli/commit/72bfb2aa8431343c0c471c01bae3649a085027a1#diff-674def7291619b0e03c5123973f4c44121cedce1c70907f77b562eec1f72da46) row `2723`)?

As I don't think Plex will offer a new call or option for a existing call in the near future to get the HDR information directly.

JonnyWong16:
Yes, even if we are only filtering for HEVC content.

herby2212:
Alright, I will close this PR until there is a more efficient way.

Issue #1405, opened 2021-04-10
Graph of Most Concurrent Streams over time


mlippert:
**Is your feature request related to a problem? Please describe.**
The _Most Concurrent Streams_ card on the home page provides some useful information, but not what I really want which is a view of _Most Concurrent Streams_ over time.

**Describe the solution you'd like**
I would like graphs of _Most Concurrent Streams_ similar to the graphs of _Play Counts_. I don't think the resolution of _Concurrent Streams_ needs to be finer grained than by hour. This would enable 3 graphs over a time span such as last 30 days:
- _Daily Most Concurrent Streams_
- _Most Concurrent Streams by day of week_
- _Most Concurrent Streams by hour of day_

The card on the home page also presents most concurrent transcodes, direct streams and direct plays. It might be useful on the _Daily Most Concurrent Streams_ graph to include lines for these as well, putting 4 lines on the graph. I do not think these need to be broken out in the _day of week_ or _hour of day_ column graphs.

**Additional context**
Graphs of _most concurrent streams_ over time broken out like this would give me an idea of how much the capabilities (cpu/ram) of my plex server are utilized/needed and when. It would also help provide an idea of when the most upload bandwidth is needed.

There was a prior request https://github.com/Tautulli/Tautulli/issues/184 which I think is where the home page most concurrent streams card came from.

PR #2046, opened 2023-04-15
Concurrent Streams per Day Graph

herby2212:
## Description

This Pull Request features a new graph called **Daily concurrent stream count by stream type** under the `Stream Type` section of the Graphs page showing the count of concurrent streams for the different stream types per day.

This feature is based on the corresponding home card and Issue #1405 .

This graph ignores the y-axis (play count/duration) property and the user filter.

### Screenshot

image

## Type of Change

## Checklist


plexpy/graphs.py
```diff
@@ -854,6 +854,103 @@ def get_total_plays_per_stream_type(self, time_range='30', y_axis='plays', user_
                   'series': [series_1_output, series_2_output, series_3_output]}
         return output
 
+    def get_total_concurrent_streams_per_stream_type(self, time_range='30'):
+        monitor_db = database.MonitorDatabase()
+
+        time_range = helpers.cast_to_int(time_range) or 30
+        timestamp = helpers.timestamp() - time_range * 24 * 60 * 60
+        
+        def calc_most_concurrent(result):
```
herby2212, line 836: Could be moved to `helpers` and be used both in `graphs` and `datafactory`. Beeing the only difference the `title` property in the `datafactory` version.

herby2212: Decided against moving it in `helpers` in favor of optimizing the local version by removing not required informations (https://github.com/Tautulli/Tautulli/pull/2046/commits/6e9fedee3b8d2cd1a321f95bdd607eda48e1e76d).

herby2212:
Ready for Review.

plexpy/webserve.py
```diff
@@ -2519,6 +2519,42 @@ def get_plays_by_stream_type(self, time_range='30', y_axis='plays', user_id=None
...
+                         {"name": "Direct Stream", "data": [...]},
+                         {"name": "Transcode", "data": [...]}
+                         ]
+                     }
+            ```
+        """
+
+        graph = graphs.Graphs()
+        result = graph.get_total_concurrent_streams_per_stream_type(time_range=time_range)
+        logger.debug(result)
```
JonnyWong16:
```suggestion

```

JonnyWong16 (changes requested):
I wonder if this graph should respect the user filter (note changes in #2090) so you can check concurrent streams per user.

Also the "Total" on the graph tooltip can be misleading. Example, I had 3 concurrent streams (2 x transcode + 1 x direct stream) and the direct play was on the same day at a different time. The graph tooltip says "Total: 4" which is technically true (I had 4 streams on that day), but not useful. I'm don't have a good way to reconcile this yet. Maybe modifying the tooltip in some way?

image

image

plexpy/webserve.py
herby2212: **User Filter**
With the change introduced in #2090 I agree that in can be included now although it will need a change in the logic.

**Tooltip**
Instead of the `Total: 4` it would be more useful to display the `max count of concurrent streams on per day` like already done on the start page.

JonnyWong16:
What if there is a fourth line in the graph for the max concurrent per day? Then four values in the tooltip, one for each line (no total).

herby2212:
> What if there is a fourth line in the graph for the max concurrent per day? Then four values in the tooltip, one for each line (no total).

This is what I proposed 4 days ago and I'm already working on as it seems like a good solution for the graph.

plexpy/graphs.py
```diff
@@ -1169,15 +1265,17 @@ def get_stream_type_by_top_10_users(self, time_range='30', y_axis='plays', user_
 
         return output
 
-    def _make_user_cond(self, user_id):
+    def _make_user_cond(self, user_id, cond_prefix=None):
```
JonnyWong16: Make the default argument `'AND'` so you don't need to check for `None`.
```suggestion
    def _make_user_cond(self, user_id, cond_prefix='AND'):
```

plexpy/graphs.py
```diff
@@ -22,7 +22,7 @@
 
 import arrow
 import datetime
-
+import itertools
```
JonnyWong16: Import is unused.
```suggestion
```

plexpy/webserve.py
```diff
@@ -2549,6 +2549,43 @@ def get_plays_by_stream_type(self, time_range='30', y_axis='plays', user_id=None
...
+                user_id (str):          Comma separated list of user id to filter the data
+
+            Returns:
+                json:
+                    {"categories":
+                        ["YYYY-MM-DD", "YYYY-MM-DD", ...]
+                     "series":
+                        [{"name": "Direct Play", "data": [...]}
+                         {"name": "Direct Stream", "data": [...]},
+                         {"name": "Transcode", "data": [...]}
```
JonnyWong16: Add 4th series.
```suggestion
                         {"name": "Transcode", "data": [...]},
                         {"name": "Max. Concurrent Streams", "data":  [...]}
```

JonnyWong16 (changes requested):
The new tooltip and 4th series is much better.

Just a few minor changes and it's good to merge. Also set a color for the "Max. Concurrent Streams" series (`graphs.html` line 330) (maybe the same blue as Live TV?).

plexpy/graphs.py
herby2212: Good point.

JonnyWong16 (approved):

Issue #1756, opened 2022-06-01
"Who has watched this" list/grid


milesgb:
### What is your feature request?

On the screen with the episode list for a single season, I would love to see a grid with episode numbers on columns and usernames on rows. Each cell would be a check mark if the user has watched the episode. This will help me decide when I can remove content.

On screens for single files, like movies, this could just be a username list with checkmarks.

Maybe nice to have.. some summary info on the whole library screen to show % users watched against each movie?

PR #2047, opened 2023-04-15
Show All Users in User Stats (Who has watched this Overview)

herby2212:
## Description

This Pull Request adds the option to show what users haven't shown a e.g. movie, episode, series,... yet by adding an option in the user stats card of the items info page. This option is only visible for users with admin rights and disabled by default. It can be enabled via a new checkbox.

This feature is based on the idea of Issue #1756 and provides those informations without adding a completely new UI element and using the existing one.

### Screenshot

**Disabled _(by default)_**
tautulliShowAllUsers_0

**Enabled**
tautulliShowAllUsers_1

## Type of Change

## Checklist


herby2212:
Ready for Review.

herby2212:
Updated to last nightly release to resolve merge conflicts. Everything works and is ready for a review.

A open point would be the option to store the setting for the user, but I don't think that it is required right now.

JonnyWong16:
I'm not convinced on the use case or need for this addition. It seems pointless to show a bunch of zeros. If a user isn't on the list then you already know they have 0 plays.

herby2212:
> I'm not convinced on the use case or need for this addition. It seems pointless to show a bunch of zeros. If a user isn't on the list then you already know they have 0 plays.

I guess the intention, which is also the benefit of it, is to get a quick overview/relation of how many users have watched the content. Without having to open a different site to get the amount of users you have.
It won't benefit many people , but also has no negative effect for others users as its disabled by default, only takes up minor space and fits from the style to the element.

A alternative to displaying the different users would be to provide, when enabled, the following as stats: `ratio of users watched this element`,` ratio of users who didn't watch this element`, `list of users who watched this element` (already existing in normal Tautulli)
image
I wouldn't call this design final, but it should display the idea.

This would be easy to understand and be benefical for nearly everyone and also fullfil the requested issue.

PR #2096, opened 2023-07-02
Fix simultaneous streams per IP not behaving as expected with IPv6

pooley182:
## Description

Because IPv6 typically does not use NAT the IPv6 address identified by Tautulli is unique to the streaming device rather than the external address of a users internet connection, like we're used to seeing with IPv4.
When limiting streams by IP this causes issues as each IPv6 devices on the same network will be treated as unique IP's.

ISP's typically lease out a /64 CIDR block to a home internet connection, this change uses the network address of a given /64 block when comparing IP addresses for identifying concurrent streams per IP address.

## Type of Change

## Checklist

### DIFF
--- a/plexpy/notification_handler.py
+++ b/plexpy/notification_handler.py
@@ -194,7 +194,6 @@ def notify_conditions(notify_action=None, stream_data=None, timeline_data=None, 
                         ip_addresses.append(helpers.get_ipv6_network_address(s['ip_address']))
                     if helpers.ip_type(s['ip_address']) == 'IPv4':
                         ip_addresses.append(s['ip_address'])
-                logger.debug("IP addresses: %s" % ip_addresses)
                 evaluated = len(Counter(ip_addresses)) >= plexpy.CONFIG.NOTIFY_CONCURRENT_THRESHOLD
             else:
                 evaluated = len(user_sessions) >= plexpy.CONFIG.NOTIFY_CONCURRENT_THRESHOLD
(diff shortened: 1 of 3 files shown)


plexpy/notification_handler.py
```diff
@@ -187,7 +188,13 @@ def notify_conditions(notify_action=None, stream_data=None, timeline_data=None,
...
 
             if plexpy.CONFIG.NOTIFY_CONCURRENT_BY_IP:
-                evaluated = len(Counter(s['ip_address'] for s in user_sessions)) >= plexpy.CONFIG.NOTIFY_CONCURRENT_THRESHOLD
+                ip_addresses = []
+                for s in user_sessions:
+                    if helpers.ip_type(s['ip_address']) == 'IPv6':
+                        ip_addresses.append(helpers.get_ipv6_network_address(s['ip_address']))
+                    if helpers.ip_type(s['ip_address']) == 'IPv4':
+                        ip_addresses.append(s['ip_address'])
+                evaluated = len(Counter(ip_addresses)) >= plexpy.CONFIG.NOTIFY_CONCURRENT_THRESHOLD
```
JonnyWong16: I think this can be a `set()` instead.

```suggestion
                ip_addresses = set()
                for s in user_sessions:
                    if helpers.ip_type(s['ip_address']) == 'IPv6':
                        ip_addresses.add(helpers.get_ipv6_network_address(s['ip_address']))
                    elif helpers.ip_type(s['ip_address']) == 'IPv4':
                        ip_addresses.add(s['ip_address'])
                evaluated = len(ip_addresses) >= plexpy.CONFIG.NOTIFY_CONCURRENT_THRESHOLD
```

JonnyWong16 (changes requested):

JonnyWong16 (approved):

PR #2166, opened 2023-09-25
Fix Plex bug where bandwidth is falsely reported as 10Gbps and quality as 0.064Mbps

itschasa:
## Description
Plex falsely reports the bandwidth and quality of transcodes as 10Gbps and 0.064Mbps respectively, when using "Convert (Maximum)" on a web browser (it's very possible this happens on other devices, but I can't test).

image

I added a helper function to detect and attempt to fix the bug using data from the video bitrate.

I multiplied the bitrate by 1.5 to get the bandwidth. It's not going to be perfectly accurate, and it never will be unless Plex fix the bug themselves, but I believe it's a good enough increase to account for spikes in bitrate during playback.

### Screenshot

Before:

image

After:

image

### Issues Fixed or Closed
Couldn't find any from a quick look, but I could be wrong.

## Type of Change
Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/plexpy/helpers.py
+++ b/plexpy/helpers.py
@@ -1793,3 +1793,24 @@ def get_ipv6_network_address(ip: str) -> str:
     if cidr_pattern.match(plexpy.CONFIG.NOTIFY_CONCURRENT_IPV6_CIDR):
         cidr = plexpy.CONFIG.NOTIFY_CONCURRENT_IPV6_CIDR
     return str(ip_network(ip+cidr, strict=False).network_address)
+
+
+def clean_plexapi_sessions(session: dict) -> dict:
+    "Cleans Plex session data, removing the 10Gbps bandwidth/0.064Mbps quality bug that appears when video transcoding sometimes."
+
+    if isinstance(session, dict):
+        # Reliable way to identify if a session is bugged.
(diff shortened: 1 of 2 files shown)


JonnyWong16:
The reserved bandwidth actually provides useful information if you are using bandwidth restrictions on your server. It tells you how much of your internet upload speed setting is being reserved and 10Gbps would indicate that something went wrong and your entire upload is reserved and no further streams can be added. Overriding that value in the UI hides this useful information making it seem like you have excess unreserved bandwidth available. I don't like seeing the 10Gbps either, but I don't think hiding/changing it's value is the right solution.

I'm pretty sure the 10Gbps is a default value when media isn't analyzed properly in Plex. Sometimes it can be fixed by analyzing your media manually or making sure your server has run it's deep analysis task on the media.

itschasa:
Mm, I see, what about changing it so the bandwidth is unchanged, but the quality profile displayed is changed? As 0.064 Mbps is definitely not accurate.

JonnyWong16:
That has more to do with transcoding from HEVC to H264 and nothing to do with the bandwidth. I think the logic I am using to estimate the profile doesn't quite work when the bitrate increases from source -> stream. I haven't looked into that in much detail.

Issue #2141, opened 2023-08-29
Show total runtime of media


pete2583:
### What is your feature request?

It would be nice to list the total runtime of all media and total library size in tb similar to how Plex dash displays it. 

It would nice to have it amongst the library total files. 

Here's an image from plex dash incase you don't know how it looks. https://i.postimg.cc/NjKKWfbM/received-738267144774203.jpg

herby2212:
The total runtime of a library is already visible as `Global Stats` when selecting a library.
The total size in TB is currently not visible, but you could get it by using the exporter configured like so:
![image](https://github.com/Tautulli/Tautulli/assets/12448284/7180f990-dd3b-40c1-9060-48d8638c3c46)

The size could also be integrated as a stat or in something like this: https://github.com/Tautulli/Tautulli/pull/1472

jaypi95:
I don't think the time displayed in the global stats is the total runtime of the media but rather the total time it has been played so far.

herby2212:
> I don't think the time displayed in the global stats is the total runtime of the media but rather the total time it has been played so far.

Agreed. I misunderstood the requirement there the watch time actually displays the time the media has really been watched by users as you mentioned correctly.

To get the runtime the exported again could be used by using `media.duration` in `Custom Media Info Fields`.

A possible solution would be the addition of those two stats in the `Media Info` section of the library.

PR #2186, opened 2023-10-28
Library Media Stats

herby2212:
## Description

This PR adds the `total media runtime` and `total file size` as indicators to several sections of the library.

The table header of the **Media Info** section of the library now directly displays those indicators.

If `Calculate Total File Size` is enabled in the config the duration aka. runtime is now included for each item. This column, like the other ones, an be shown/hidden.
image

The **Libraries** overview also includes those as fields in two new configurable columns.
image

For this a new method `get_library_media_stats` got added to the API which returns the total size, the total duration aka. runtime and total storage of a library.
This data gets cached into a new file `media_stats_<section_id>` to allow for faster loading times. For this the caching methods for saving and loading have been rebuild to a generic version.

After first updating to this version, when enabled in the config, the file size and duration will be recalculated and saved to the cache for the items.

The API documentation got updated.

### Issues Fixed or Closed

- Implements #2141.

## Type of Change

## Checklist


JonnyWong16:
My idea was to decouple the media info from the total. We can get the total directly from Plex without adding up all the individual items. This means the total can be shown regardless of the file sizes setting. 

Refer to the Plex API library here:

https://github.com/pkkid/python-plexapi/blob/a3af807c6bb2f4bb5a150f4ddff05314a30093db/plexapi/library.py#L462C17-L462C17

This also opens up the option to add the totals to columns on the Libraries table (cache to database `library_sections` on libraries list refresh).

herby2212:
> My idea was to decouple the media info from the total. We can get the total directly from Plex without adding up all the individual items. This means the total can be shown regardless of the file sizes setting.
> 
> Refer to the Plex API library here:
> 
> https://github.com/pkkid/python-plexapi/blob/a3af807c6bb2f4bb5a150f4ddff05314a30093db/plexapi/library.py#L462C17-L462C17

Interesting. This would save both on loading time and also on complexity, but would mean to replace the calculation logic entirely. It would also allow to make a new call which can be used internally for the total, but also provided via the API.

I will start working on that, but keep the visual changes so far made as those still seem fitting to me.

Since when is this call supported by Plex? Do we need to check for a special plex server version.

> This also opens up the option to add the totals to columns on the Libraries table (cache to database `library_sections` on libraries list refresh).

Good call👍

JonnyWong16:
> Since when is this call supported by Plex? Do we need to check for a special plex server version.

Over 3 years when Plex Dash was released.

Might need to verify that the API endpoint isn't Plex Pass only because Plex Dash is Plex Pass only. I don't think it is.

herby2212:
Ready for review. I updated the main message to reflect the new changes.

I also decided to keep the addition of the duration in the file size calculation per item as it seems like a logical addition and doesn't increase the runtime as no additional call is made.
The fields containing the total indicators use the new call including caching.

Did you verify if the calls for the properties are Plex Pass only? Does the python-plexapi include a field or indication for premium required functions/calls, if not this would seem like a good addition.

JimbooMan:
**I dont know if this is a bug or not. But i wanted to sort out which of my movies that have been watched. If i choose a movie that i know have been seen by me and others the all time watch time is 0. Why is that? I installed Tautulli using the simple way with the EXE file. Is there anything else that i need?
Will it only count from the day that i installed Tautulli?** 

Tautulli

herby2212:
> I dont know if this is a bug or not. But i wanted to sort out which of my movies that have been watched. If i choose a movie that i know have been seen by me and others the all time watch time is 0. Why is that? I installed Tautulli using the simple way with the EXE file. Is there anything else that i need?

Please use the issue section of github to report a problem or ask for support in the future  (https://github.com/Tautulli/Tautulli/issues) as this discussion is only regarding the content of the PR.
> Will it only count from the day that i installed Tautulli?

Yes as Tautulli tracks informations far more detailed then Plex does and so needs to run to gather this data.

herby2212:
@JonnyWong16 Still ready for review. Any timeframe the review will be done?

JonnyWong16:
I will get to this soon. I've been busy.

plexpy/libraries.py
```diff
@@ -143,6 +145,44 @@ def has_library_type(section_type):
     result = monitor_db.select_single(query=query, args=args)
     return bool(result)
 
+def get_library_media_stats(section_id=None, refresh=False):
```
JonnyWong16, line 148: Save to the database `library_sections` table instead of caching to a file. Add two new columns `total_storage` and `total_duration`. `total_size` is already the `count` column in the database.

You can keep the `total_size` in the returned dict for the API.

plexpy/libraries.py
```diff
@@ -409,18 +449,23 @@ def get_datatables_list(self, kwargs=None, grouping=None):
             else:
                 library_art = item['library_art']
 
+            library_media_stats = get_library_media_stats(item['section_id'])
```
JonnyWong16, line 452: Once added to the database, just add the columns to the query above.

JonnyWong16 (changes requested):
Revert the cache changes. It's making this PR very difficult to review.

PR #2301, opened 2024-04-08
feat: add support for bypassing gosu in container image

onedr0p:
## Description

Hi 👋🏼 

This adds support in the Docker image to bypass gosu, the reason being for this is I have Tautulli deployed within Kubernetes and it provides this functionality out of the box by using `fsGroupChangePolicy` as described in the Kubernetes docs on [security contexts](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/).

The change here would allow me to set an env var of `TAUTULLI_DOCKER_GOSU` to `false` and use the Kubernetes native feature of changing ownership of files.

There could be a few different ways of handling this so please let me know if this is acceptable or if you would like this to be handled a different way.

Thanks 🎉 

### Screenshot

Include screenshots if the changes are UI-related.

### Issues Fixed or Closed

- Fixes #(issue)

## Type of Change

## Checklist


start.sh
```diff
@@ -1,16 +1,20 @@
...
 
-    find /config \! \( -uid $(id -u tautulli) -gid $(id -g tautulli) \) -print0 | xargs -0r chown tautulli:tautulli
+        find /config \! \( -uid $(id -u tautulli) -gid $(id -g tautulli) \) -print0 | xargs -0r chown tautulli:tautulli
 
-    echo "Running Tautulli using user tautulli (uid=$(id -u tautulli)) and group tautulli (gid=$(id -g tautulli))"
-    exec gosu tautulli "$@"
+        echo "Running Tautulli using user tautulli (uid=$(id -u tautulli)) and group tautulli (gid=$(id -g tautulli))"
+        exec gosu tautulli "$@"
+    else
+        exec tautulli "$@"
```
JonnyWong16: Don't you still need to create a `tautulli` user?

onedr0p: That doesn't matter, using kubernetes or docker compose people can override the user and group anyways as long as they have their permissions on the volume set correctly.

JonnyWong16: I meant that this command will fail because there is no `tautulli` user inside the container.

onedr0p: Actually I see what you mean the tautulli in the command is the user.

onedr0p: Updated.

onedr0p:
@JonnyWong16 

There's a way I found out to do this without the PR, while it's not as straight forward as using the env that would be implemented here the below does work

```yaml
            command: ["python", "Tautulli.py"]
            args: [
              "--config", "/config/config.ini",
              "--datadir", "/config",
              "--port", "80"
            ]
```

PR #2543, opened 2025-04-29
Allow users to set config values through environment variables.

### LINKED ISSUE #2309: Allow users to set config values through environment variables on startup
### Is your feature request related to a problem?

I just started migrating my Tautulli instance and the new instance is hosted on docker and managed through docker compose. In front of the new instance is a reverse proxy and hence I want to be able to specify the `http_root` variable through docker compose. 

### What is your feature request?

The industry standard way of doing this is by exposing such configuration using environment variables. It would be very usefull if Tautulli could support something like that.

### Are there any workarounds?

My current solution was configuring everything in docker compose and reverse proxy as if Tautulli was ready for reverse proxy. I then launched it for the first time using `docker-compose up`, then stopping it as soon as it had populated the data folder. I then manually edited the `config.ini` file with the `http_root` value I wanted. Then I started Tautulli back up again.

### Additional Context

_No response_

komuw:
## Description
- Allow users to set config values through environment variables. 

### Screenshot
N/A

### Issues Fixed or Closed

- Fixes #2309

## Type of Change

## Checklist


plexpy/config.py
```diff
@@ -487,13 +487,27 @@ def check_setting(self, key):
...
 
+    def _from_env(self, key, definition_type):
+        """ Get key from environment variables, if it exists """
+        val = os.environ.get(key)
+        try:
+            if val:
+                return definition_type(val)
+        except Exception:
+            return None
+
```
komuw, line 524: 1. There's a possibility of an environment variable from other programs clashing with one from Tautulli. An environment variable like `HTTP_PORT` could also be in use by another program apart from Tautulli. A way to fix that is to require that all environment variables for Tautulli to have a predetermined prefix. Something like `TAUTULLI_` so the previous environ variable would be `TAUTULLI_HTTP_PORT`. Should we require a prefix?
2. In the current PR, I've implemented it in a way that environment variables have higher precedence than config file values. I think that's the right call, since someone who has set the environment variable really wants those to take effect. But I'm willing to be persuaded otherwise.
3. I wanted to add one or two unit tests, but I cannot find any tests in the project. Is there a convention on where to put them?

JonnyWong16: 1. Yes, prefix with `TAUTULLI_`.
2. Agreed, environmental variables should have higher precedence.
3. There are no tests in the project.

komuw: > 1. Yes, prefix with TAUTULLI_. 

I've added the prefix in https://github.com/Tautulli/Tautulli/pull/2543/commits/4610bf37692d5c1629f9d3d90999f3db5d224b46   

Thanks for the review.

JonnyWong16:
If they environment variable and the config file value are different, should the environment variable get saved to the config? Or should the config file value be kept and the environment variable only applies to the current instance?

JonnyWong16:
I rewrote the handling of environment variables so they don't overwrite the config file. If you try to save a setting that is defined as an environment variable it will just keep the existing value in the config file. This might cause confusion because it will seem like the settings page is not working.

komuw:
> If they environment variable and the config file value are different, should the environment variable get saved to the config? Or should the config file value be kept and the environment variable only applies to the current instance?

That's a good observation that I hadn't taken into account.
> I rewrote the handling of environment variables so they don't overwrite the config file. If you try to save a setting that is defined as an environment variable it will just keep the existing value in the config file. This might cause confusion because it will seem like the settings page is not working.

That makes sense. Thank you.  


Is there anything else you'd like me to help with on the PR?

PR #2655, opened 2026-03-20
Add rating field to get_home_stats API response

jma1ice:
## Description

`top_movies`, `popular_movies`, `top_tv`, `popular_tv`, and `last_watched` in datafactory.py now pull `shm.rating` and pass it to the get_home_stats API response. Local docker image made after changes to ensure they work and nothing else breaks

### Screenshot

No UI changes

### Issues Fixed or Closed

N/A

## Type of Change

- [✓] New feature (non-breaking change which adds functionality)

## Checklist

- [✓] My code follows the style guidelines of this project
- [✓] I have performed a self-review of my own code
- [✓] I have commented my code, particularly in hard-to-understand areas
- [✓] I have added or updated the docstring for new or existing methods (info was added into the webserve example response)

### DIFF
--- a/plexpy/datafactory.py
+++ b/plexpy/datafactory.py
@@ -399,8 +399,9 @@ class DataFactory(object):
             if stat == 'top_movies':
                 top_movies = []
                 try:
-                    query = "SELECT sh.id, shm.full_title, shm.year, sh.rating_key, shm.thumb, sh.section_id, " \
-                            "shm.art, sh.media_type, shm.content_rating, shm.labels, sh.started, shm.live, shm.guid, " \
+                    query = "SELECT sh.id, shm.full_title, shm.year, sh.rating_key, shm.thumb, " \
+                            "sh.section_id, shm.art, sh.media_type, shm.content_rating, shm.rating, " \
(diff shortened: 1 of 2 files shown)


JonnyWong16:
Why isn't it added to `top_music` and `popular_music` as well?

jma1ice:
> Why isn't it added to `top_music` and `popular_music` as well?

Are there ratings on audio libraries? I don't have music, only audiobooks and didn't see ratings on those so I assumed it wouldn't work but didn't double check the Tautulli database to see. I can add these in today if they will work!

JonnyWong16:
Even if the value is blank, I think it's better to include it just so the API response is consistent.

jma1ice:
Definitely makes sense! I'll get that added in as soon as I'm home later today and report back

jma1ice:
> Even if the value is blank, I think it's better to include it just so the API response is consistent.

Alright, the response is present now in `top_music` and `popular_music` as well!

JonnyWong16 (approved):
Thanks.

jma1ice:
Thank you! Verified API call working for rating in newsletterr

PR #2708, opened 2026-05-14
fix: replace newlines with Markdown hard line breaks in Gotify messages

### LINKED ISSUE #2702: Gotify: Single line breaks in notification text are not rendered correctly due to hardcoded text/markdown
### Describe the Bug

The built-in Gotify notification agent hardcodes contentType: text/markdown in the extras field. This causes Gotify clients to render the message body as Markdown, which collapses single line breaks (entered via the Tautulli notification text editor) into spaces.

If you enter f.e. the following two lines with simply pressing Enter for new line:
Test1
Test2

it becomes: Test1 Test2

Markdown has to be used instead of text/plain, because Gotify can only work with poster images when using markdown.

The root cause stems from [notifiers.py](https://github.com/Tautulli/Tautulli/blob/master/plexpy/notifiers.py):
https://github.com/Tautulli/Tautulli/blob/d7b8147944c745c9dd566ac7839ed21f941d6d1d/plexpy/notifiers.py#L1713-L1722
Tautulli stores newlines in notification templates as 0x0A (\n). In Markdown, a single \n is not a hard line break — it is collapsed into a space by the renderer. The correct Markdown hard line break requires two trailing spaces for a line break which can then be followed by newline, which isn't rendered anyways (  \n).

### Steps to Reproduce

1. Set up a Gotify notification agent in Tautulli
2. In the message body, enter two lines separated by a single line break (Enter key)
3. Send a test notification
4. Observe the output in the Gotify Web UI or Gotify for Android

### Expected Behavior

Each line break entered in the Tautulli notification editor should result in a visible line break in the received notification.

### Suggested fix:
We could simply enter two spaces before every new line, but not everybody knows this or could easily find out, why this would solve the problem. Instead we should do it automatically and invisibly:
Replace "\n" with "  \n" (two trailing spaces + newline) in the message body before sending, which is the correct Markdown syntax for a hard line break, e.g. change:
https://github.com/Tautulli/Tautulli/blob/d7b8147944c745c9dd566ac7839ed21f941d6d1d/plexpy/notifiers.py#L1720 to
`'message': body.replace('\n', '  \n'),`

This is a one-line fix with no breaking changes — text/markdown stays set (preserving the poster image feature), and line breaks work correctly in all official Gotify clients.

---
_Additionally we could set contentType to text/plain (and then not use the two trailing spaces) when "Include Poster Image" is disabled, simply for security reasons. This is of course if that's the only valid reason, why an image would be rendered. I don't know at this point, if there exists another way to f.e. include one or multiple rendered images in the message body as well. The reason for this is stated in the [Gotify documentation](https://gotify.net/docs/msgextras#definition):_

> Markdown supports rendering images via , these images will be automatically downloaded when the message is viewed. Similarly, to why remote contents are blocked by default in e-mail clients, automatically downloading remote images could be used to collect information from the user.
Also, if part of the message is interpolated from a malicious external source, the attacker could inject malformed markdown which leads to information disclosure.
It is recommended to use text/plain to reduce possible security issues when using text from external sources like f.ex. output from scripts.

### Screenshots

_No response_

### Relevant Settings

_No response_

### Tautulli Version

v2.17.1

### Git Branch

master

### Git Commit Hash

unknown

### Platform and Version

DSM 7.3 (Synology) via Docker Image (LSIO)

### Python Version

3.12.13 (main, Apr 10 2026, 13:58:11) [GCC 15.2.0]

### Browser and Version

Firefox v150.0.2

### Link to Logs

not really needed IMHO

Gilmoursa:
## Summary

Fixes #2702 — single line breaks in Gotify notification text were being collapsed into spaces.

**Root cause:** The Gotify notification agent hardcodes `contentType: text/markdown` (required to support poster images). In Markdown, a single `\n` is not a hard line break — it renders as a space. So notification text entered with simple Enter key presses would arrive as a single run-on line.

**Fix:** Replace `\n` with `  \n` (two trailing spaces + newline) in the message body before sending. This is the correct Markdown syntax for a hard line break, as suggested in the issue. `text/markdown` is kept so poster images continue to work.

## Change

`plexpy/notifiers.py` — one line in `GOTIFY.agent_notify()`:

```python
# before
'message': body,

# after
'message': body.replace('\n', '  \n'),
```

## Test plan

- [ ] Set up a Gotify notification agent in Tautulli
- [ ] Enter a message body with single line breaks (Enter key)
- [ ] Send a test notification and confirm line breaks render correctly in Gotify
- [ ] Confirm poster images still render when "Include Poster Image" is enabled

### DIFF
--- a/plexpy/notifiers.py
+++ b/plexpy/notifiers.py
@@ -1717,7 +1717,7 @@ class GOTIFY(Notifier):
                     'contentType': 'text/markdown'
                 }
             },
-            'message': body,
+            'message': body.replace('\n', '  \n'),
             'priority': self.config['priority']
         }
 


JonnyWong16:
This is already done.

aac59a6ba28770ccf955e3468149401139bbf7d1

Gilmoursa:
Closing as this has already been addressed in aac59a6 per @JonnyWong16. Thanks!

PR #2722, opened 2026-06-05
Fix empty host fallback in launch_browser

upmcplanetracker:
## Description

This PR updates the `plexpy.launch_browser` function to safely normalize an empty string (`''`) to `localhost`. 

Currently, `launch_browser` correctly normalizes wildcard addresses like `0.0.0.0` and `::`. However, if a user explicitly configures `HTTP_HOST` to an empty string (e.g., via environment variables like `TAUTULLI_HTTP_HOST=` to force native dual-stack wildcard binding via the OS as can be done in Rootless Podman Quadlets), the `launch_browser` function fails to catch it. 

Without this normalization, Tautulli attempts to format an invalid URL (e.g., `http://:8181`), which causes:
1. The startup auto-launcher to fail.
2. System tray/menu bar "Open" actions on Windows and macOS to attempt to open a broken web address. 

### What was changed?
Added `''` to the existing tuple check in `launch_browser`, ensuring that `host in ('0.0.0.0', '::', '')` safely falls back to `localhost` across all launch triggers.

### Screenshot

N/A (Backend logic change fixing broken browser redirection/launch strings; no visible UI elements modified)

### Issues Fixed or Closed

- Refers to automated feedback from GitHub Copilot on previous design exploration. I deleted my previous pull request/fork because Copilot rightly showed that it would break Tautulli (for everyone!).  I found a much more elegant and simple solution that works on my system/podman quadlet and should work for everyone.

## Type of Change

## Checklist

### DIFF
--- a/plexpy/__init__.py
+++ b/plexpy/__init__.py
@@ -404,7 +404,7 @@ def daemonize():
 
 def launch_browser(host, port, root):
     if not no_browser:
-        if host in ('0.0.0.0', '::'):
+        if host in ('0.0.0.0', '::', ''):
             host = 'localhost'
 
         if CONFIG.ENABLE_HTTPS:


JonnyWong16 (approved):
Thanks.

upmcplanetracker:
i appreciate all the hard work you do.  thanks.

PR #2728, opened 2026-06-12
Paginate datatables queries in SQL instead of Python

### LINKED ISSUE #1962: Increase performance by reducing complexity of joins
### Is your feature request related to a problem?

_No response_

### What is your feature request?

Like many people I have a large and growing history table (350k+ rows) that is becoming extremely slow and unwieldy in Tautulli. But this is not a limitation of SQLite3, as some have assumed. The issue is continuously increasing complexity of joins and costly pagination strategy.

For example on the default history page a huge query is run that has several joins and a union which end up needing to match every single row in the history table. Only once every row of a temp table is generated does pagination come into play: in fact, the entire database is practically joined together, read into RAM, and then immediately deleted/discarded with a list slice.

Without a major rework, one simple solution is to change the way pagination works.

Instead of doing this huge join on every row, you should calculate just a list of IDs of relevant records, and either use that as a subtable in a "WHERE ... IN" clause to limit the complexity of the joins, or run two queries (one to generate a list of IDs, one to actually fetch the enriched data).

Here's an example query generated by Tautulli on the main history page (which, by the way, has no instrumentation currently for tracking slow queries). This query takes nearly 20 seconds on my system.

```sql
SELECT * FROM (
  SELECT 
    session_history.reference_id, 
	session_history.id AS row_id, 
	MAX(started) AS date, 
	MIN(started) AS started, 
	MAX(stopped) AS stopped, 
	SUM(CASE WHEN stopped > 0 THEN (stopped - started) ELSE 0 END) - SUM(CASE WHEN paused_counter IS NULL THEN 0 ELSE paused_counter END) AS duration, 
	SUM(CASE WHEN paused_counter IS NULL THEN 0 ELSE paused_counter END) AS paused_counter, 
	session_history.user_id, 
	session_history.user, 
	(CASE WHEN users.friendly_name IS NULL OR TRIM(users.friendly_name) = "" THEN users.username ELSE users.friendly_name END) AS friendly_name, 
	users.thumb AS user_thumb, 
	users.custom_avatar_url AS custom_thumb, 
	platform, product, player, ip_address, machine_id, location, secure, relayed, session_history.media_type, 
	(CASE WHEN session_history_metadata.live = 1 THEN 'live' ELSE session_history.media_type END) AS media_type_live, 
	session_history_metadata.rating_key, 
	session_history_metadata.parent_rating_key, 
	session_history_metadata.grandparent_rating_key, 
	session_history_metadata.full_title, 
	session_history_metadata.title, 
	session_history_metadata.parent_title, 
	session_history_metadata.grandparent_title, 
	session_history_metadata.original_title, 
	session_history_metadata.year, 
	session_history_metadata.media_index, 
	session_history_metadata.parent_media_index, 
	session_history_metadata.thumb, 
	session_history_metadata.parent_thumb, 
	session_history_metadata.grandparent_thumb, 
	session_history_metadata.live, 
	session_history_metadata.added_at, 
	session_history_metadata.originally_available_at, 
	session_history_metadata.guid, 
	MAX((CASE WHEN (view_offset IS NULL OR view_offset = "") THEN 0.1 ELSE view_offset * 1.0 END) /(CASE WHEN (session_history_metadata.duration IS NULL OR session_history_metadata.duration = "") THEN 1.0 ELSE session_history_metadata.duration * 1.0 END) * 100) AS percent_complete, 
	session_history_media_info.transcode_decision, 
	COUNT(*) AS group_count, 
	GROUP_CONCAT(session_history.id) AS group_ids, 
	NULL AS state, 
	NULL AS session_key 
	
  FROM session_history 
  LEFT OUTER JOIN users ON session_history.user_id = users.user_id 
  JOIN session_history_metadata ON session_history.id = session_history_metadata.id 
  JOIN session_history_media_info ON session_history.id = session_history_media_info.id   
 
  GROUP BY session_history.reference_id 
  
  UNION 
    SELECT 
	  NULL AS reference_id, 
	  NULL AS row_id, 
	  started AS date, 
	  started, stopped, 
	  SUM(CASE WHEN stopped > 0 THEN (stopped - started) ELSE (strftime("%s", "now") - started) END) - SUM(CASE WHEN paused_counter IS NULL THEN 0 ELSE paused_counter END) AS duration, 
	  SUM(CASE WHEN paused_counter IS NULL THEN 0 ELSE paused_counter END) AS paused_counter, 
	  user_id, user, 
	  (CASE WHEN friendly_name IS NULL OR TRIM(friendly_name) = "" THEN user ELSE friendly_name END) AS friendly_name, 
	  NULL AS user_thumb, 
	  NULL AS custom_thumb, 
	  platform, product, player, ip_address, machine_id, location, secure, relayed, media_type, 
	  (CASE WHEN live = 1 THEN 'live' ELSE media_type END) AS media_type_live, 
	  rating_key, parent_rating_key, grandparent_rating_key, full_title, title, parent_title, grandparent_title, original_title, year, media_index, parent_media_index, thumb, parent_thumb, grandparent_thumb, live, added_at, originally_available_at, guid, 
	  MAX((CASE WHEN (view_offset IS NULL OR view_offset = "") THEN 0.1 ELSE view_offset * 1.0 END) / (CASE WHEN (duration IS NULL OR duration = "") THEN 1.0 ELSE duration * 1.0 END) * 100) AS percent_complete, 
	  transcode_decision, 
	  NULL AS group_count, 
	  NULL AS group_ids, 
	  state, session_key 
	FROM sessions  
	GROUP BY session_key
)
ORDER BY date COLLATE NOCASE DESC
```

The above query results in 350k rows, which are then sliced down to *25*. From what I can tell you could greatly reduce the complexity of the joins (to constant time!) with a WHERE .. IN like this:

```sql

-- In the first temp table (before the union)

  WHERE session_history.id IN (
    select id from (
      SELECT id,max(started) as date from session_history GROUP BY session_history.reference_id ORDER BY date COLLATE NOCASE DESC LIMIT 25
	)
  )
```

This drops the query from 20 seconds to 900ms for me. But 900ms is still far slower than it needs to be. The cost is in the aggregate column `max(started)`. Since this is used so frequently, it would be hugely advantageous to store a special flag for whether a session_history entry is the most recent to have been started. It is easy to update that flag on prior records for a given reference_id at insert time, and makes the search query instantaneous.

I also noticed something a bit confusing: the two slow joins are joining session_history, session_history_metadata, and session_history_media_info. At least on my install these tables all have the same row count, and don't serve any apparent relational purpose. I could see perhaps the existence of the other two tables as a way to reduce rows and track media items atomically, but that is not the case with Tautulli right now; a row is inserted into all 3 every time. So why do these other tables exist at all? Maybe they could be merged together into session_history if there's no relational value, reducing the need for joins.

I want to make a PR to help here but the DataTables + MonitorDatabase combo contain so much custom SQL-generation code that I'm not sure the best strategy to assist. But I hope this analysis is useful. There's no need for people to suffer multi-second pageloads. These queries should be near-instant and SQLite3 is fine -- also, switching to another backend like MySQL/Postgres is not going to improve anything by much anyway.

### Are there any workarounds?

_No response_

### Additional Context

_No response_

stephenp:
## Description

Fixes the slow history table queries described in #1962 by applying pagination in SQL instead of Python.

Previously, `DataTables.ssp_query` ran the full joined/grouped query with no `LIMIT`, materialized every matching row into Python dicts, computed `recordsFiltered` with `len()`, and then sliced out the requested page with a Python list slice. On large databases (the issue reports 350k+ history rows taking ~20s per page load) this meant joining, grouping, and converting the entire history table into Python just to display 25 rows.

This change, confined to `ssp_query` in `plexpy/datatables.py`:

1. Appends `LIMIT ? OFFSET ?` (bound parameters) to the query so SQLite only returns the requested page.
2. Computes recordsFiltered with a separate `SELECT COUNT(*)` over the same inner query instead of counting a fully materialized result set.

Notes:
- No frontend or caller changes; the tables already use DataTables server-side processing and send start/length with every request; they're just applied in SQL now. The response contract (`recordsFiltered` / `recordsTotal` / `data` / `draw`) is unchanged.
- All tables that go through `ssp_query` get the same fix for free: history, users, user IP addresses, login logs, libraries, notification/newsletter logs, and exports.
- Side fix: page length "all" (`length=-1`) previously dropped the last row due to the slice arithmetic (`filtered[start:start + (-1)]);` SQLite's `LIMIT -1` handles it correctly.
- `start`/`length` are now read with defaults and cast to int, so a malformed request degrades gracefully instead of raising.
- SQLite still evaluates the `GROUP BY` internally for the count, so there's room for a further optimization (e.g. the `WHERE id IN` (subquery) rewrite suggested in #1962) as a follow-up — but this removes the dominant cost (materializing every row into Python) with a minimal, low-risk diff.

### Screenshot

N/A

### Issues Fixed or Closed

- Fixes #1962 

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas


putnam:
It would be good to post the complete query generated for the history table with this change applied. By the looks of it this is simply adding a limit/offset to one query, and not tackling the bigger issues like the aggregate column usage (which can be precomputed and cached instead) and the use of massive joins. When you're selecting from one big table, obviously a limit/offset is the solution for pagination. But when you are doing several joins you want to reduce the working set as much as possible, which is where the WHERE...IN clause comes into play. I'm not seeing that here. Is this tested on a big history table?

Another problem with low level pagination like this is that you can accidentally starve higher level filters of rows, and end up with fewer rows in the pages than you want. But I can't tell from just looking at the PR whether that could be the case. You'd have to investigate every calling site and see how they're using this abstraction layer.

I'm not the owner of this repo but I would also say it is good and ethical to be up-front and direct with your use of AI both for commits themselves and the text body of your comments/issues/PRs.

stephenp:
That's a fair callout. 

I'm more comfortable in FE land and hoped I'd be able to leverage what BE I know to contribute in a meaningful way (with AI), but I should have disclosed AI was used.

I did test this with my own DB of 35,369 history rows compared to your 572k, so mine isn't quite as large as yours.

I had Claude generate a test db of ~600k rows and saw some improvements on the history page query, but I just noticed there was a slight addition of latency for smaller db users that I didn't catch the first time.

I agree that this is definitely **not** the big win you were hoping for when you opened the issue, but was hoping to at least chip away at _some_ of the issue with a really small, safe, and scoped change. Writing complex SQL / refactoring huge queries is definitely not in my wheelhouse and even with AI I wouldn't be comfortable contributing something I wasn't confident in. The `WHERE...IN` refactor is above my paygrade.

I'll defer to the BE experts from here on out on whether to actually merge it since it introduces some new latency for smaller DB users.

This PR does seem to improve large DB query times, but I think even my own DB performance wouldn't have _that_ noticeable of a change. 

The following before/after through the actual `get_datatables_history` path (best of 5, warm) was generated / formatted with Claude:

### Real database — 35k history rows / 25.7k groups

  | Scenario | `master` (full fetch + Python slice) | This PR (SQL `LIMIT`/`OFFSET`) | Change |
  | --- | ---: | ---: | :---: |
  | Page 1, grouped _(default view)_ | 0.414s | **0.288s** | **−30%** |
  | Page 1, ungrouped | 0.515s | **0.321s** | **−38%** |
  | Page 800 (deep offset) | 0.410s | 0.403s | ~0% |
  | Search `"the"` (13,451 matches) | 0.327s | 0.384s | +17% :warning: |

### Synthetic database — 600k history rows / 540k groups

  | Scenario | `master` (full fetch + Python slice) | This PR (SQL `LIMIT`/`OFFSET`) | Change |
  | --- | ---: | ---: | :---: |
  | Page 1, grouped _(default view)_ | 6.937s | **2.568s** | **−63%** |
  | Page 1, ungrouped | 7.609s | **2.634s** | **−65%** |
  | Page 800 (deep offset) | 6.952s | **3.051s** | **−56%** |
  | Search `"the"` (204,949 matches) | 4.250s | **3.986s** | **−6%** |

_Edit: Forgot to address your other concerns from the issue thread._

I had Claude audit every `ssp_query` caller (`get_datatables_history`, `get_user_unique_ips`, `get_user_login`, and the `users/libraries/notify/newsletter/exports` tables), but none of them filter or aggregate the result set in Python. 

They all iterate the returned page, reshape each row for display, and pass `filteredCount`/`totalCount` straight through. No caller that depends on receiving the full set, the only thing that changes is that the unused rows are never built into Python dicts and discarded.

JonnyWong16 (approved):
Thanks.

This may not be the full database optimization, but at least it's a start in the right direction.

PR #2769, opened 2026-08-02
Require matching guid when grouping non-live history

vefilippo:
## Summary
- require matching `guid` when grouping non-live history rows
- keep the existing live-TV grouping behavior unchanged
- prevent `regroup_history()` from re-merging unrelated items when legacy `rating_key` collisions exist

## Problem
`regroup_history()` currently replays `group_history()` across all stored history rows. For non-live media, `group_history()` groups by `user_id + rating_key` and reuses the previous row's `reference_id`.

On long-lived databases that have survived Plex/library migrations, historical `rating_key` collisions can exist across different media items. When that happens, regrouping can merge unrelated plays back together even after `reference_id` has been repaired.

Verified example from my database:
- `Modern Family - Wine Weekend`
- `NCIS: Origins - Operazione Tramonto`
- both stored with `rating_key = 8908`
- different `guid` values
- `Regroup play history` merged them again because the current logic trusted `rating_key`

## Fix
For non-live media, this patch still narrows candidates by `user_id + rating_key`, but only reuses the previous `reference_id` when the previous row and current row also share the same `guid`.

That keeps normal resume/regroup behavior for the same item while avoiding false merges caused by stale `rating_key` collisions in legacy history.

## Validation
- syntax check with `py_compile`
- verified against the failing `rating_key = 8908` case from my backup DB
- with this patch, the bad pair no longer qualifies for regroup because the `guid` values differ

### DIFF
--- a/plexpy/activity_processor.py
+++ b/plexpy/activity_processor.py
@@ -484,6 +484,7 @@ class ActivityProcessor(object):
     def group_history(self, last_id, session, metadata=None):
         new_session = prev_session = None
         prev_watched = None
+        session_guid = metadata['guid'] if metadata else session.get('guid')
 
         db = database.MonitorDatabase()
 
@@ -512,9 +513,15 @@ class ActivityProcessor(object):
                 prev_watched = False
 
         else:
-            # Check if we should group the session, select the last two rows from the user
-            query = "SELECT id, rating_key, view_offset, reference_id FROM session_history " \
-                    "WHERE id <= ? AND user_id = ? AND rating_key = ? ORDER BY id DESC LIMIT 2 "
+            # Check if we should group the session, select the last two rows from the user.
+            # For non-live media, also compare guid so regrouping won't merge unrelated
+            # history rows that happen to share a stale rating_key.
+            query = "SELECT session_history.id, session_history.rating_key, session_history.view_offset, " \
+                    "session_history.reference_id, session_history_metadata.guid " \
+                    "FROM session_history " \
+                    "LEFT OUTER JOIN session_history_metadata ON session_history.id == session_history_metadata.id " \
+                    "WHERE session_history.id <= ? AND session_history.user_id = ? " \
+                    "AND session_history.rating_key = ? ORDER BY session_history.id DESC LIMIT 2 "
 
             args = [last_id, session['user_id'], session['rating_key']]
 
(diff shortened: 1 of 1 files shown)


PIBER:
Thanks for tracking this down — the root cause analysis here is correct, and I can
confirm the bug occurs organically, not only via DB imports.

On a production server with ~111k history rows I found **308 plays** mis-grouped this
way across **234 distinct recycled rating_keys**, spanning 2018–2026. The average gap
between the old play and the new one that inherited its key was **1,392 days** (max
3,213). No import was involved — Plex simply reuses rating_keys as items are deleted
and added, so any library with regular turnover hits this continuously. In my case the
rate is ~30 rows/month and rising. Symptom is exactly as described: with
`group_history_tables = 1`, the new play renders under the old item's title, so it looks
like the play was never recorded.

## One problem with the patch as written

I believe this will disable pause/resume grouping entirely for live-logged sessions.

`group_history()` is called from `write_session_history()` at **activity_processor.py:324**,
but `session_history_metadata` is not written until **:479** — in the same function, after
the grouping call. Nothing writes that table in between (only `session_history_media_info`
at :414). And since the insert uses `keys = {'id': None}`, every call creates a brand new
row, so there is no pre-existing metadata row to find.

That means the new `LEFT OUTER JOIN session_history_metadata` returns `guid = NULL` for
`result[0]` (the row just inserted). With the new condition:

```python
not session['live'] and prev_session['guid'] and new_session['guid']
and prev_session['guid'] == new_session['guid'] and ...
```

`new_session['guid']` is `None` → falsy → the non-live clause is always `False` → nothing
is ever grouped during normal logging.

`regroup_history()` (:757) calls `group_history()` with `metadata=None` on rows whose
metadata already exists, so that path works correctly — which I suspect is why this
wasn't caught in testing.

## Suggested change

Take the new session's guid from the `metadata` argument, the way the live-TV branch
already does at :509. `metadata` is guaranteed non-None at the call site (it is
dereferenced as `metadata['section_id']` at :313), and falling back to the joined column
keeps `regroup_history()` working:

```python
new_session = {'id': result[0]['id'],
               'rating_key': result[0]['rating_key'],
               'view_offset': helpers.cast_to_int(result[0]['view_offset']),
               'reference_id': result[0]['reference_id'],
               'guid': (metadata['guid'] if metadata
                        else result[0]['guid'] or session.get('guid'))}

prev_session = {'id': result[1]['id'],
                'rating_key': result[1]['rating_key'],
                'view_offset': helpers.cast_to_int(result[1]['view_offset']),
                'reference_id': result[1]['reference_id'],
                'guid': result[1]['guid']}
```

The rest of the PR (the JOIN and the guid comparison in the condition) is unchanged.

I've been running exactly this variant in production since 2026-09-15. Happy to report
back after it has accumulated more hours, or to open a separate PR if that's easier.

One behavioural note worth flagging: requiring a guid match also stops grouping across
legitimate metadata-agent refreshes, where the same media acquires a new guid. On my
server that affects ~371 rows historically. The effect is minor — an occasional resumed
movie shows as two rows instead of one — but it is a real trade-off rather than a pure
win, and may deserve a mention in the changelog.

PR #2772, opened 2026-08-04
Tautulli Remote Relay Support

TheMeanCanEHdian:
## Description

OneSignal's free plan drops to under 1,000 monthly active users on October 1st, 2026, with no
grandfathering. Tautulli Remote is well above that, so on that date notifications stop working for most
users of the app.

This adds a second delivery transport for the Tautulli Remote App notifier. It is a push relay that
forwards notifications to Firebase Cloud Messaging, and on to Apple's push service through FCM for iOS
devices. The relay is a Cloudflare Worker, open source under GPL v3 at
[Tautulli/Tautulli-Remote-Relay](https://github.com/Tautulli/Tautulli-Remote-Relay). The official instance runs at `relay.tautulliremote.com`.

Both transports run side by side. The choice is made per device, from the value stored in
`mobile_devices.push_token`, and not from any version comparison:

| `push_token` | Meaning | Transport |
| --- | --- | --- |
| `NULL` or empty | registered by an app version predating the relay | OneSignal, as before |
| `'push-disabled'` | new app, notifications turned off or token unavailable | neither, excluded from the device list |
| an FCM token | new app with notifications enabled | relay |

There is no flag day and no coordinated upgrade. A server running this can notify old and new app versions
at the same time, and users move over as they update the app. Removing the OneSignal path is a separate
change for after October 1st and is not part of this PR.

### What changed

<details>
<summary>New and changed functions</summary>

New

| | |
| --- | --- |
| `mobile_app.relay_device_id()` | SHA-256 prefix of a push token, for use wherever the token itself cannot be logged or shown |
| `mobile_app.validate_push_token()` | Checks a token against the relay: valid, dead, or indeterminate |
| `mobile_app.validates_remotely()` | Whether a device holds a credential worth a validation call |
| `mobile_app.set_official_from_delivery()` | Updates `official` from the result of a send |
| `mobile_app.revalidate_devices()` | Startup repair pass for unvalidated devices, on its own thread |
| `mobile_app._revalidate_devices()` | Its body: filters, paces, validates |
| `TAUTULLIREMOTEAPP._via_relay()` | Picks the transport for a device; shared by the sender and the config screen |
| `TAUTULLIREMOTEAPP._trim_to_relay_limit()` | Fits the payload to the relay's cap, or reports that it cannot |
| `TAUTULLIREMOTEAPP._send_via_relay()` | Posts to the relay and maps each response onto a log line |

Changed

| | |
| --- | --- |
| `mobile_app.set_official()` | Takes `push_token` and validates whichever credential was registered; an indeterminate result no longer clears a validated device |
| `mobile_app.add_mobile_device()` | Accepts and stores `push_token` |
| `mobile_app.get_mobile_device_config()` | Exposes `relay_device_id` to the device modal instead of the raw credential |
| `TAUTULLIREMOTEAPP.agent_notify()` | Selects a transport per device; drops a notification that cannot be trimmed to fit |
| `TAUTULLIREMOTEAPP.get_devices()` | Lists devices holding either credential, and excludes the disabled sentinel |
| `TAUTULLIREMOTEAPP._return_config_options()` | Names the delivery path, and keeps the OneSignal data collection notice for a device still delivered through it |
| `api2.register_device()` | Optional `push_token`, blacklisted from the logs on receipt |
| `request.request_response2()` | Defaults `verify` rather than assigning it, so a caller can require verification |
| `plexpy.upgrade()` and `initialize()` | Device revalidation moves off the one-shot upgrade flag and runs on every startup |

</details>

`plexpy/notifiers.py`

The Tautulli Remote App notifier picks a transport per device, using the relay when a push token is present
and is not the disabled sentinel. `_send_via_relay()` posts the existing notification envelope to the
configured relay URL and maps its responses onto separate log messages, so a dead token, a rate limit and
an oversized payload are each diagnosable instead of all reading "notification failed".

`get_devices()` now selects devices holding either credential, so relay-registered devices appear in the
notifier's device list. Devices carrying the disabled sentinel are excluded by the same query, which is why
they never reach a delivery attempt on either transport.

Relay payloads are capped at 3800 bytes, so `_trim_to_relay_limit()` shortens the notification before it is
sent. The budget is computed for the envelope actually being sent, since the encrypted form pays for base64
expansion and the unencrypted form does not. The body is shortened first and the subject only if emptying
the body was not enough, because the subject is the notification's headline. If the payload still does not
fit, the notification is dropped with an error rather than sent in a state the relay will refuse.

Three logging behaviours are worth calling out. A dead push token reports at error level once and at debug
afterwards: clearing the official flag only drops the device from the notifier configuration list, not from
an agent already saved against it, so a phone the user has stopped using would otherwise repeat the same
error on every notification. A rate limit says which of the two it is, because the relay attaches rate
limit details only for the daily fair use cap; a burst refusal clears in a minute, the cap does not clear
until it resets, and neither is retried. And every failure names the device the same way, as friendly name
and hashed id, since the id on its own is a token digest that nothing resolves to a phone.

The relay requests set `allow_redirects=False` and `verify=True`. The relay only ever answers 200, 410, 413,
429 or 5xx, so a redirect is always anomalous, and a 307 or 308 would re-send the push token to whatever
`Location` names, because the token travels in the request body rather than a header.

`plexpy/mobile_app.py`

`validate_push_token()` checks a token against the relay, mirroring the existing `validate_onesignal_id()`.
`set_official()` validates whichever credential the device actually registered with. An indeterminate
result, such as a rate limit or a relay or FCM outage, no longer clears a device that has previously
validated. Clearing it would drop the device out of the notifier's device list until the user re-registered
by hand, with nothing to indicate why.

`revalidate_devices()` repairs devices left unvalidated by an earlier outage. It runs on every startup, off
the startup thread, and skips devices that are already validated or that have no registration to check, so
a server where nothing needs repair does no work and logs nothing. It is paced at one device every two
seconds to stay inside the relay's per-IP limit on validation.

`relay_device_id()` returns the first sixteen characters of a push token's SHA-256, following the existing
`hash_pms_uuid()`. The push token is blacklisted from the logs, so without it the relay path had no way to
say which device a line referred to. The relay records the same value and Tautulli Remote displays it, so a
user's report, this log and the relay's own records can be lined up.

`plexpy/request.py`

`request_response2()` defaults `verify` instead of assigning it. It previously overwrote whatever a caller
passed, so a caller could not require verification for a single request. No existing call site passes
`verify`, so nothing else changes. The relay request uses it: `VERIFY_SSL_CERT` exists for the user's own
Plex server, and the relay is a third party host receiving a credential, so it should not follow that
setting.

`plexpy/__init__.py`

Adds `mobile_devices.push_token`, using the existing try-`SELECT` / except-`ALTER` dbcheck pattern.

Device revalidation moves out of `upgrade()` and into `initialize()`, unconditionally. It was gated behind
`CONFIG.UPGRADE_FLAG == 0`, a one-shot that has shipped set since v2.10.0, so on an existing install the
pass never ran and a device left unvalidated by an outage was never repaired. `UPGRADE_FLAG` now has no
readers; its config key is left in place.

`plexpy/api2.py`

`register_device` accepts an optional `push_token` parameter. Existing callers are unaffected. The token is
added to the logger blacklist on receipt, so it does not appear in Tautulli's logs.

`plexpy/config.py`

Adds `REMOTE_APP_PUSH_URL` under Advanced, defaulting to the official relay. Anyone running a fork of the
app with their own Firebase project can point their server at their own relay.

HTML

Three wording changes, all of them transport-neutral rather than relay-specific:

- `mobile_device_config.html`: the device field is labelled "Notification Device ID" instead of "OneSignal
  Device ID", and shows the hashed identifier described above, falling back to `onesignal_id` for devices
  that only have one. It replaces a field that displayed the raw credential, which an admin has no use for.
- `mobile_devices_table.html`: the failed-validation tooltip reads "Notification Registration Validation
  Failed" rather than naming OneSignal, since either credential can now fail to validate.
- `settings.html`: the "must not be blocked in Pi-hole" note names `relay.tautulliremote.com`, which is
  what someone configuring a DNS filter actually needs.

### What did not change

Notification encryption is untouched. The same AES-256-GCM envelope with a PBKDF2-derived key is used for
both transports. The relay forwards an opaque payload and cannot read it, in the same way OneSignal could
not.

The notifier's configuration and behaviour are unchanged for existing users. There are no settings to
migrate and nothing a server admin needs to do.

The OneSignal code path is untouched, and remains the default for every device already registered.

### Testing

Tautulli has no test suite, so all of this was run against throwaway harnesses rather than committed tests.

- A contract suite of 38 checks against the real `agent_notify()` and a real SQLite database, covering
  transport selection, the schema migration, `official` handling on each response code, and the encrypted
  payload decrypting with a PBKDF2 key derived from the device token.
- 14 end-to-end checks of a real Tautulli against a real relay over HTTP. These confirm that a failed
  delivery does not consume a device's quota, that an oversized payload is trimmed and accepted while an
  untrimmed one is refused, and that a validation failure during an outage returns indeterminate rather
  than invalid.
- Verified on an Android device end to end: registration, `official` set from a live relay validation, and
  an encrypted notification delivered, decrypted and displayed.
- Verified that the schema migration runs cleanly against an existing `tautulli.db`.

A later review pass over the branch produced the size-limit, redirect, certificate and logging changes
described above, and those were checked separately: the trim behaviour across both envelope shapes and the
refusal path, redirect refusal against a local 308 responder using the vendored `requests`, the `verify`
resolution for every combination of `VERIFY_SSL_CERT` and caller argument, the transport predicate against
every shape of device row `get_devices()` can return, and the notifier help text rendered for each
combination of cryptography availability and device transport.

### Issues Fixed or Closed

N/A

## Type of Change

## Checklist

### DIFF
--- a/data/interfaces/default/settings.html
+++ b/data/interfaces/default/settings.html
@@ -2091,7 +2091,7 @@ Rating: {rating}/10   --> Rating: /10
                     or manually enter the connection info and device token into the app settings. This window will automatically close once device registration is successful.
                 </p>
                 <p class="help-block">
-                    Note: relay.tautulliremote.com must not be blocked (e.g. in Pi-hole) for device registration.
+                    Note: Tautulli must be able to reach relay.tautulliremote.com (e.g. not blocked in Pi-hole) to validate the device and deliver notifications.
                 </p>
                 <label>QR Code</label>
                 <div id="api_qr_code"></div>
--- a/data/interfaces/default/settings.html
+++ b/data/interfaces/default/settings.html
@@ -2091,7 +2091,7 @@ Rating: {rating}/10   --> Rating: /10
                     or manually enter the connection info and device token into the app settings. This window will automatically close once device registration is successful.
                 </p>
                 <p class="help-block">
-                    Note: The Tautulli push relay (and OneSignal.com, for app versions before the relay) must not be blocked (e.g. in Pi-hole) for device registration.
+                    Note: The Tautulli Remote relay (and OneSignal.com, for app versions before the relay) must not be blocked (e.g. in Pi-hole) for device registration.
                 </p>
                 <label>QR Code</label>
                 <div id="api_qr_code"></div>
--- a/data/interfaces/default/settings.html
+++ b/data/interfaces/default/settings.html
@@ -2091,7 +2091,7 @@ Rating: {rating}/10   --> Rating: /10
                     or manually enter the connection info and device token into the app settings. This window will automatically close once device registration is successful.
                 </p>
                 <p class="help-block">
-                    Note: The Tautulli Remote relay (and OneSignal.com, for app versions before the relay) must not be blocked (e.g. in Pi-hole) for device registration.
+                    Note: relay.tautulliremote.com must not be blocked (e.g. in Pi-hole) for device registration.
                 </p>
                 <label>QR Code</label>
                 <div id="api_qr_code"></div>
--- a/data/interfaces/default/settings.html
+++ b/data/interfaces/default/settings.html
@@ -2091,7 +2091,7 @@ Rating: {rating}/10   --> Rating: /10
                     or manually enter the connection info and device token into the app settings. This window will automatically close once device registration is successful.
                 </p>
                 <p class="help-block">
-                    Note: OneSignal.com must not be blocked (e.g. in Pi-hole) for device registration.
+                    Note: The Tautulli push relay (and OneSignal.com, for app versions before the relay) must not be blocked (e.g. in Pi-hole) for device registration.
                 </p>
                 <label>QR Code</label>
                 <div id="api_qr_code"></div>
--- a/plexpy/__init__.py
+++ b/plexpy/__init__.py
@@ -260,9 +260,6 @@ def initialize(config_file):
         notifiers.blacklist_logger()
         mobile_app.blacklist_logger()
 
-        # Repair any device left unvalidated by an earlier outage
-        mobile_app.revalidate_devices()
-
         # Check if Tautulli has a uuid
         if CONFIG.PMS_UUID == '' or not CONFIG.PMS_UUID:
             logger.debug("Generating UUID...")
@@ -546,6 +543,9 @@ def start():
         # Start background notification thread
         notification_handler.start_threads(num_threads=CONFIG.NOTIFICATION_THREADS)
         notifiers.check_browser_enabled()
+
+        # Repair any device left unvalidated by an earlier outage
+        mobile_app.revalidate_devices()
 
         # Schedule newsletters
         newsletter_handler.NEWSLETTER_SCHED.start()
--- a/plexpy/__init__.py
+++ b/plexpy/__init__.py
@@ -260,6 +260,9 @@ def initialize(config_file):
         notifiers.blacklist_logger()
         mobile_app.blacklist_logger()
 
+        # Repair any device left unvalidated by an earlier outage
+        mobile_app.revalidate_onesignal_ids()
+
         # Check if Tautulli has a uuid
         if CONFIG.PMS_UUID == '' or not CONFIG.PMS_UUID:
             logger.debug("Generating UUID...")
@@ -2880,11 +2883,6 @@ def dbcheck():
 
 def upgrade():
     logger.info("Checking if the configurastion upgrades are required...")
-
-    if CONFIG.UPGRADE_FLAG == 0:
-        mobile_app.revalidate_onesignal_ids()
-        CONFIG.UPGRADE_FLAG = 1
-        CONFIG.write()
 
     logger.info("Configuration upgrade complete.")
 
--- a/plexpy/__init__.py
+++ b/plexpy/__init__.py
@@ -261,7 +261,7 @@ def initialize(config_file):
         mobile_app.blacklist_logger()
 
         # Repair any device left unvalidated by an earlier outage
-        mobile_app.revalidate_onesignal_ids()
+        mobile_app.revalidate_devices()
 
         # Check if Tautulli has a uuid
         if CONFIG.PMS_UUID == '' or not CONFIG.PMS_UUID:
--- a/plexpy/__init__.py
+++ b/plexpy/__init__.py
@@ -811,7 +811,7 @@ def dbcheck():
         "CREATE TABLE IF NOT EXISTS mobile_devices (id INTEGER PRIMARY KEY AUTOINCREMENT, "
         "device_id TEXT NOT NULL UNIQUE, device_token TEXT, device_name TEXT, "
         "platform TEXT, version TEXT, friendly_name TEXT, "
-        "onesignal_id TEXT, last_seen INTEGER, official INTEGER DEFAULT 0)"
+        "onesignal_id TEXT, push_token TEXT, last_seen INTEGER, official INTEGER DEFAULT 0)"
     )
 
     # tvmaze_lookup table :: This table keeps record of the TVmaze lookups
@@ -2398,6 +2398,15 @@ def dbcheck():
                 "SELECT device_id FROM mobile_devices WHERE official > 0").fetchall():
             c_db.execute("UPDATE mobile_devices SET platform = ? WHERE device_id = ?",
                          ["android", device_id])
+
+    # Upgrade mobile_devices table from earlier versions
+    try:
+        c_db.execute("SELECT push_token FROM mobile_devices")
+    except sqlite3.OperationalError:
+        logger.debug("Altering database. Updating database table mobile_devices.")
+        c_db.execute(
+            "ALTER TABLE mobile_devices ADD COLUMN push_token TEXT"
+        )
 
     # Upgrade notifiers table from earlier versions
     try:
--- a/plexpy/api2.py
+++ b/plexpy/api2.py
@@ -375,7 +375,7 @@ class API2(object):
         return data
 
     def register_device(self, device_id='', device_name='', platform=None, version=None,
-                        friendly_name='', onesignal_id=None, min_version='', **kwargs):
+                        friendly_name='', onesignal_id=None, push_token=None, min_version='', **kwargs):
         """ Registers the Tautulli Remote App.
 
             ```
@@ -388,6 +388,7 @@ class API2(object):
                 version (str):            The version of the app
                 friendly_name (str):      A friendly name to identify the mobile device
                 onesignal_id (str):       The OneSignal id for the mobile device
+                push_token (str):         The push notification token for the mobile device
                 min_version (str):        The minimum Tautulli version supported by the mobile device, e.g. v2.5.6
 
             Returns:
@@ -433,7 +434,9 @@ class API2(object):
             return
 
         ## TODO: Temporary for backwards compatibility, assume device_id is onesignal_id
-        if device_id and onesignal_id is None:
+        # App versions that register a push token always send onesignal_id explicitly,
+        # so this only applies to older apps that predate both.
+        if device_id and onesignal_id is None and push_token is None:
             onesignal_id = device_id
 
         result = mobile_app.add_mobile_device(device_id=device_id,
@@ -442,7 +445,8 @@ class API2(object):
                                               platform=platform,
                                               version=version,
                                               friendly_name=friendly_name,
-                                              onesignal_id=onesignal_id)
+                                              onesignal_id=onesignal_id,
+                                              push_token=push_token)
 
         if result:
             self._api_msg = 'Device registration successful.'
@@ -707,6 +711,8 @@ General optional parameters:
                 logger._BLACKLIST_WORDS.add(kwargs['device_id'])
             if kwargs.get('onesignal_id'):
                 logger._BLACKLIST_WORDS.add(kwargs['onesignal_id'])
+            if kwargs.get('push_token'):
+                logger._BLACKLIST_WORDS.add(kwargs['push_token'])
 
         result = None
         logger.api_debug('Tautulli APIv2 :: API called with kwargs: %s' % kwargs)
--- a/plexpy/mobile_app.py
+++ b/plexpy/mobile_app.py
@@ -224,7 +224,11 @@ def set_official(device_id, onesignal_id, push_token=None):
                            + where,
                            args=[official, platform, device_id])
     except Exception as e:
-        logger.warn("Tautulli MobileApp :: Failed to set official flag for device: %s." % e)
+        # No device identifier here: this runs for both transports, and a
+        # OneSignal device has nothing loggable to name it by. Its device id and
+        # OneSignal id are both blacklisted, which is why the OneSignal
+        # validation lines carry no identifier either.
+        logger.warn("Tautulli MobileApp :: Failed to set official flag: %s." % e)
         return
 
 
@@ -246,7 +250,8 @@ def set_official_from_delivery(device, official):
         db.action("UPDATE mobile_devices SET official = ? WHERE device_id = ?",
                   args=[official, device['device_id']])
     except Exception as e:
-        logger.warn("Tautulli MobileApp :: Failed to set official flag for device: %s." % e)
+        logger.warn("Tautulli MobileApp :: Failed to set official flag for device %s: %s."
+                    % (relay_device_id(device['push_token']), e))
 
 
 def set_last_seen(device_token=None):
--- a/plexpy/mobile_app.py
+++ b/plexpy/mobile_app.py
@@ -272,8 +272,9 @@ def validate_push_token(push_token):
 
     logger.info("Tautulli MobileApp :: Validating push token for device %s", device_id)
     try:
+        # A 307 or 308 would re-post the push token to wherever Location points.
         r = requests.post('%s/v1/validate' % plexpy.CONFIG.REMOTE_APP_PUSH_URL.rstrip('/'),
-                          headers=headers, json=payload, timeout=10)
+                          headers=headers, json=payload, timeout=10, allow_redirects=False)
         status_code = r.status_code
         logger.info("Tautulli MobileApp :: Push token validation for device %s returned status code %s",
                     device_id, status_code)
--- a/plexpy/mobile_app.py
+++ b/plexpy/mobile_app.py
@@ -15,6 +15,7 @@
 #  You should have received a copy of the GNU General Public License
 #  along with Tautulli.  If not, see <http://www.gnu.org/licenses/>.
 
+import hashlib
 import requests
 import threading
 import time
@@ -35,6 +36,19 @@ _PUSH_DISABLED = 'push-disabled'
 _REVALIDATE_INTERVAL = 2
 
 TEMP_DEVICE_TOKENS = {}
+
+
+def relay_device_id(push_token):
+    """Return the identifier the push relay knows a device by.
+
+    The push token is a credential and is blacklisted from the logs, so it
+    cannot be used to say which device a log line is about. The relay derives
+    this same digest, and Tautulli Remote shows it on its data dump page, so
+    one string ties a user's report, this log and the relay's records together.
+    """
+    if not push_token or push_token == _PUSH_DISABLED:
+        return 'none'
+    return hashlib.sha256(push_token.encode()).hexdigest()[:16]
 
 
 def set_temp_device_token(token=None, remove=False, add=False, success=False):
@@ -137,6 +151,11 @@ def get_mobile_device_config(mobile_device_id=None):
 
     if result['push_token'] == _PUSH_DISABLED:
         result['push_token'] = ''
+
+    # Shown in place of the token itself, which is a credential the admin has no
+    # use for. This is the value the relay records and the app displays, so it is
+    # the one worth quoting in a bug report.
+    result['relay_device_id'] = relay_device_id(result['push_token']) if result['push_token'] else ''
 
     return result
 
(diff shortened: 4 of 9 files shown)


data/interfaces/default/settings.html
```diff
@@ -2091,7 +2091,7 @@ <h4 class="modal-title">Register Tautulli Remote App</h4>
                     or manually enter the connection info and device token into the app settings. This window will automatically close once device registration is successful.
                 </p>
                 <p class="help-block">
-                    Note: OneSignal.com must not be blocked (e.g. in Pi-hole) for device registration.
+                    Note: relay.tautulliremote.com must not be blocked (e.g. in Pi-hole) for device registration.
```
JonnyWong16: Is `relay.tautulliremote.com` required for device registration, or just for notification registration? If the relay is blocked the device will still register with Tautulli but will have `push-disabled` ID.

plexpy/__init__.py
```diff
@@ -260,6 +260,9 @@ def initialize(config_file):
         notifiers.blacklist_logger()
         mobile_app.blacklist_logger()
 
+        # Repair any device left unvalidated by an earlier outage
+        mobile_app.revalidate_devices()
```
JonnyWong16: Move inside `start()` function after Tautulli is fully initialized.

plexpy/__init__.py
```diff
@@ -2872,11 +2884,6 @@ def dbcheck():
 def upgrade():
     logger.info("Checking if the configurastion upgrades are required...")
 
-    if CONFIG.UPGRADE_FLAG == 0:
```
JonnyWong16, line 2875: `UPGRADE_FLAG` no longer used elsewhere in Tautulli. It can be removed from the config file (`config.py`).

plexpy/api2.py
```diff
@@ -433,7 +434,9 @@ def register_device(self, device_id='', device_name='', platform=None, version=N
             return
 
         ## TODO: Temporary for backwards compatibility, assume device_id is onesignal_id
-        if device_id and onesignal_id is None:
+        # App versions that register a push token always send onesignal_id explicitly,
+        # so this only applies to older apps that predate both.
+        if device_id and onesignal_id is None and push_token is None:
```
JonnyWong16, line 439: Can this backwards compatibility be removed by now? Or leave as-is until after Oct. 1 when we do OneSignal cleanup.

plexpy/mobile_app.py
```diff
@@ -210,9 +259,61 @@ def validate_onesignal_id(onesignal_id):
...
+    headers = {'Content-Type': 'application/json'}
+    payload = {'token': push_token}
+
+    device_id = relay_device_id(push_token)
+
+    logger.info("Tautulli MobileApp :: Validating push token for device %s", device_id)
+    try:
+        # A 307 or 308 would re-post the push token to wherever Location points. The relay
+        # is verified regardless of VERIFY_SSL_CERT.
+        r = requests.post('%s/v1/validate' % plexpy.CONFIG.REMOTE_APP_PUSH_URL.rstrip('/'),
```
JonnyWong16:
```suggestion
        r = requests.post(f'{plexpy.CONFIG.REMOTE_APP_PUSH_URL.rstrip('/')}v1/validate',
```

plexpy/mobile_app.py
```diff
@@ -210,9 +259,61 @@ def validate_onesignal_id(onesignal_id):
...
 
 
-def revalidate_onesignal_ids():
-    for device in get_mobile_devices():
-        set_official(device['device_id'], device['onesignal_id'])
+def validate_push_token(push_token):
+    if push_token == _PUSH_DISABLED:
+        return 2
+
+    headers = {'Content-Type': 'application/json'}
```
JonnyWong16: `Content-Type` header is automatically added by `requests` when using `requests.post(..., json=payload)`.

plexpy/mobile_app.py
```diff
@@ -210,9 +259,61 @@ def validate_onesignal_id(onesignal_id):
...
+
+    device_id = relay_device_id(push_token)
+
+    logger.info("Tautulli MobileApp :: Validating push token for device %s", device_id)
+    try:
+        # A 307 or 308 would re-post the push token to wherever Location points. The relay
+        # is verified regardless of VERIFY_SSL_CERT.
+        r = requests.post('%s/v1/validate' % plexpy.CONFIG.REMOTE_APP_PUSH_URL.rstrip('/'),
+                          headers=headers, json=payload, timeout=10,
+                          allow_redirects=False, verify=True)
```
JonnyWong16: `requests` default is `verify=True`.

plexpy/notifiers.py
```diff
@@ -4009,11 +4009,154 @@ class TAUTULLIREMOTEAPP(Notifier):
...
+                         % (self.NAME, encoded_length() - budget))
+            return False
+
+        logger.warn("Tautulli Notifiers :: %s notification %s to fit the notification size limit."
+                    % (self.NAME, ' and '.join(trimmed)))
+        return True
+
+    def _send_via_relay(self, device, data):
+        payload = {'token': device['push_token'],
+                   'platform': device['platform'] or 'android',
```
JonnyWong16: Why fallback to `'android'`?

plexpy/notifiers.py
```diff
@@ -4009,11 +4009,154 @@ class TAUTULLIREMOTEAPP(Notifier):
...
+                   'data': data}
+
+        url = '%s/v1/notify' % plexpy.CONFIG.REMOTE_APP_PUSH_URL.rstrip('/')
```
JonnyWong16:
```suggestion
        url = f'{plexpy.CONFIG.REMOTE_APP_PUSH_URL.rstrip('/')}/v1/notify'
```

plexpy/notifiers.py
```diff
@@ -4009,11 +4009,154 @@ class TAUTULLIREMOTEAPP(Notifier):
...
+                if encoded_length() <= budget:
+                    low = mid
+                else:
+                    high = mid - 1
+
+            if low == len(text):
+                return None
+
+            plaintext_data[key] = text[:low] + '...'
+            return "%s shortened from %s to %s characters" % (key, len(text), low)
```
JonnyWong16:
```suggestion
            return f"{key} shortened from {len(text)} to {low} characters"
```

plexpy/notifiers.py
```diff
@@ -4009,11 +4009,154 @@ class TAUTULLIREMOTEAPP(Notifier):
...
+            logger.error("Tautulli Notifiers :: {name} notification to device {device} failed: the notification is "
+                         "too large to deliver. Shorten the notification text.".format(
+                             name=self.NAME, device=device_label))
+
+        else:
+            logger.error("Tautulli Notifiers :: {name} notification to device {device} failed: the push relay "
+                         "returned status code {status}.".format(
+                             name=self.NAME, device=device_label, status=response.status_code))
+
+        logger.debug("Tautulli Notifiers :: Request response: {}".format(request.server_message(response, True)))
```
JonnyWong16: Do we need to log the response here?

plexpy/notifiers.py
```diff
@@ -4009,11 +4009,154 @@ class TAUTULLIREMOTEAPP(Notifier):
...
+                logger.error("Tautulli Notifiers :: {name} notification to device {device} failed: the daily "
+                             "notification limit has been reached. Further notifications to this device will "
+                             "fail until the limit resets at {reset}.".format(
+                                 name=self.NAME, device=device_label,
+                                 reset=rate_limits.get('resetsAt', 'the next reset')))
+            else:
+                logger.error("Tautulli Notifiers :: {name} notification to device {device} failed: notifications "
+                             "are being rate limited. Retry after {retry} seconds.".format(
+                                 name=self.NAME, device=device_label,
+                                 retry=response.headers.get('Retry-After', 'unknown')))
```
JonnyWong16, line 4153: There is no auto-retry implemented. The notification is just dropped.

JonnyWong16 (changes requested):

plexpy/__init__.py
TheMeanCanEHdian: Moved into `start()` in 77f3bd12, after `notifiers.check_browser_enabled()`. It spawns its own thread internally, so it won't hold up startup.

plexpy/__init__.py
TheMeanCanEHdian: Removed in 77f3bd12 — note the change is in `plexpy/config.py`, so it won't show as outdated on this thread. `UPGRADE_FLAG` was the only remaining reference in the tree once the gate here was dropped.

plexpy/mobile_app.py
TheMeanCanEHdian: Dropped the dict and the `headers=` kwarg in 77f3bd12.

plexpy/notifiers.py
TheMeanCanEHdian: Removed in 77f3bd12 — every branch above it already logs a specific message, so it was duplicate debug output.

plexpy/notifiers.py
TheMeanCanEHdian: Reworded in 77f3bd12 to say the notification was dropped, and to report `Retry-After` verbatim rather than implying anything retries.

JonnyWong16: Can we implement an auto-retry? Could it be implemented in some generic method that could be extended to other notification agents?

plexpy/mobile_app.py
TheMeanCanEHdian: Removed in 7be202c2, along with the comment.

One thing worth flagging: the `verify=True` on the relay send in `notifiers.py` is load-bearing and I have left it in place. That call goes through `request.request_response2`, which applies `VERIFY_SSL_CERT` via `setdefault`, so without the explicit argument a user who disabled certificate verification for their own server would also disable it when POSTing the push token to the relay.

plexpy/mobile_app.py
TheMeanCanEHdian: Applied in principle in 06e9aaef, with two adjustments.

The suggestion drops the path separator. `rstrip('/')` strips the trailing slash from the configured URL and the `/` was previously supplied by the format string, so moving the interpolation into the f-string without re-adding it yields:

```
https://relay.tautulliremote.comv1/validate
```

Every validation would then fail — quietly, since `validate_push_token` catches the exception and returns `-1`, so devices simply never reach `official = 1`. The companion suggestion on `notifiers.py` has the slash, so this looks like a slip rather than intent.

I also switched the outer quotes to double. `f'{... .rstrip('/')}...'` nests same-type quotes, which is PEP 701 and requires Python 3.12+; on 3.11 and below it is a `SyntaxError` at import, so Tautulli would not start. I could not find an existing use of that construct in the tree or a declared minimum version, so I kept it compatible — happy to use single quotes throughout if 3.12 is already the floor.

Landed as:

```python
r = requests.post(f"{plexpy.CONFIG.REMOTE_APP_PUSH_URL.rstrip('/')}/v1/validate",
```

JonnyWong16: The `/` and `"` were just my typos.

plexpy/notifiers.py
TheMeanCanEHdian: Applied in 476cf749.

plexpy/notifiers.py
TheMeanCanEHdian: Applied in 476cf749, with the outer quotes doubled per your note about the typo:

```python
url = f"{plexpy.CONFIG.REMOTE_APP_PUSH_URL.rstrip('/')}/v1/notify"
```

plexpy/notifiers.py
TheMeanCanEHdian: The fallback mirrors `set_official`, which backfills a NULL platform with `'android'` when validation succeeds (`mobile_app.py:199`, applied via `coalesce(platform, ?)` at `:209`).

But it is the wrong default. The relay branches on platform when building the FCM message: `ios` gets an `apns` block with `mutable-content: 1`, `android` gets a data-only message. An iOS device sent as `android` therefore never runs its notification service extension, so the encrypted payload is never decrypted and the user gets nothing — with no error logged anywhere.

In practice this is rare. The app has always sent `platform`, before and after the relay migration, and it is non-nullable on the device side, so a NULL only appears for rows registered outside the official app or edited by hand. But when it does happen, the current behaviour is a silent failure on precisely the platform that cannot tolerate the guess.

Options as I see them:

1. **Leave as-is.** Never sends an invalid request, accepts silent iOS breakage in the rare case.
2. **Drop the fallback** and send `device['platform']` as-is. The relay already requires it and returns 400 (`"platform" must be one of: "android", "ios"`). Smallest possible diff — though now that the response-body debug log is gone, the log would show only `returned status code 400`, so the reason would not be visible.
3. **Guard before sending** — skip and log something specific, e.g. "unknown platform for device X, re-register it in Tautulli Remote". No wasted relay call, nothing counted against the device fair-use quota, and the operator gets an actionable message.

Either 2 or 3 also argues for changing `mobile_app.py:199` to leave NULL as NULL, so the guess is not persisted to the database — `revalidate_devices()` calls `set_official` for every device on startup (`:313`), so today a legacy row would get `'android'` written in permanently.

My inclination is 3 plus the `set_official` change, but 2 is the smallest diff if you would rather keep it minimal. Happy either way.

data/interfaces/default/settings.html
TheMeanCanEHdian: Registration doesn't need the relay. `add_mobile_device` upserts the row before `set_official` runs in a thread, so a blocked relay leaves `official = -1`, which renders as "Notification Registration Validation Failed". That warning is the only registration-time symptom, so the note stays rather than being dropped.

`push-disabled` isn't network-related — the app sends it only when consent or notification permission is missing. If it can't reach FCM, registration throws instead. And unlike OneSignal, the relay is contacted by Tautulli rather than the device.

Reworded to:

> Note: Tautulli must be able to reach relay.tautulliremote.com (e.g. not blocked in Pi-hole) to validate the device and deliver notifications.

It could be dropped entirely if you'd prefer.

plexpy/api2.py
TheMeanCanEHdian: Leaving as-is until the Oct 1 cleanup.

It still fires for devices on the pre-v3 app, which used the OneSignal player ID as its device identifier. Current app versions send `onesignal_id` explicitly (`'onesignal-disabled'`) and a `push_token`, so the condition cannot be reached from anything recent. Removing it now would leave a legacy device re-registering with `onesignal_id = NULL` and no way to identify it.

After Oct 1 that stops mattering, since those devices cannot be delivered to through OneSignal either way, and this can go along with `validate_onesignal_id` and the column itself.

data/interfaces/default/settings.html
JonnyWong16: That wording is better.

plexpy/notifiers.py
TheMeanCanEHdian: **TL;DR:** the relay's burst limit is a 60-second window, so an in-place retry either gets refused again or blocks a notification worker for a full minute. Delivering a retried notification would need deferred re-queueing; the generic hook itself is ~50-60 lines and changes nothing for agents that don't opt in.

The blocker isn't the retry loop, it's the shape of the limit.

The relay's burst limiter is 30 requests per 60 seconds and it advertises `Retry-After: 60`, so there's no short window to ride out. An in-place retry either sleeps less than that and gets refused again — adding latency and a wasted request before dropping the notification anyway — or sleeps the full minute. `NOTIFICATION_THREADS` defaults to 2, so the second option takes out half of notification delivery for a minute, when the queue is busiest. `api2.py:506` also calls `notification_handler.notify()` synchronously and shapes the result into the API response, so a retry on that path hangs an HTTP request rather than a background thread.

Actually delivering a retried notification means deferred re-delivery — re-queueing with a due time, or an APScheduler job. That needs somewhere to hold the pending notification across a restart, a decision about staleness (a "now playing" notification arriving a minute late), and `notify_log` semantics for an attempt that hasn't resolved yet.

The generic part is cheap on its own: a `RETRY_ATTEMPTS` class attribute defaulting to 0, a `retry_delay(response)` classifier returning `None` by default, and a wrapper around `request_response2` that `make_request` and `_send_via_relay` both route through. Around 50-60 lines, no change to `make_request`'s return contract, no edits at the 19 call sites, and no behaviour change for any agent that doesn't opt in.

It has to be opt-in rather than default-on because 429 doesn't mean the same thing everywhere. Pushover's 429 is a monthly quota and goes through the shared `make_request`, so a generic 429 retry would burn a user's remaining allowance faster. Telegram returns its backoff in the response body, Pushbullet uses `X-Ratelimit-Reset`, and the self-hosted targets (Gotify, ntfy, webhooks) are whatever the user runs.

None of this is driven by observed traffic — `DAILY_LIMIT` is still 0 and there's no usage data yet, so whether anyone reaches the burst limit in practice is unknown.

plexpy/notifiers.py
JonnyWong16: I like option 3 with the change in `set_official` as well.

TheMeanCanEHdian: Done in aabc4972.

`_send_via_relay` now refuses rather than guessing, and logs the device with a pointer to re-register it. The accepted values match the relay's own `PLATFORMS`, so the two agree on what is valid.

`set_official` no longer writes a platform at all — the `UPDATE` only touches `official` now. Since `revalidate_devices()` runs it for every device on startup, the guess was being persisted on every boot.

One side effect: the Android-only Priority option in the agent config is gated on `platform == 'android'`, so a legacy device with no recorded platform will no longer show it. That seems right, since the platform is genuinely unknown at that point.

JonnyWong16 (approved):

Issue #1435, opened 2021-05-10
Archiving Feature


Jerome2103:
**Is your feature request related to a problem? Please describe.**
Archiving

**Describe the solution you'd like**
An archiving option for users who no longer have access. So that you can hide them from the user list and from the history. The same features for music, series and movies that are no longer available on Plex.

**Describe alternatives you've considered**
Waiting

**Additional context**

samwiseg0:
This is already possible? If you delete a user their history is removed and they are not longer in the user list. 

They can be un-deleted if that changes. Does that cover what you are looking for? Are we missing something?

Jerome2103:
Would like to have a hidden button for users, who do not want to delete the history etc. Therefore archive and do not delete.

PR #2778, opened 2026-08-09
Add option to archive users without deleting their history

akonopcz:
## Problem

Deleting a user always purges their history. `Users.delete()` calls
`database.delete_user_history()` and sets `keep_history = 0`, so hiding a user and keeping
what they watched is currently impossible.

That leaves two options for someone who no longer has access to the server: delete them and
lose the history for good, or keep them in the users list, the history table and every
statistic forever. It is particularly awkward for users who come and go, since re-sharing
the server does not bring back what was deleted.

This is the open question from #1435. When it was reported, the answer was:

> This is already possible? If you delete a user their history is removed and they are no
> longer in the user list.

and the reporter clarified:

> Would like to have a hidden button for users, who do not want to delete the history etc.
> Therefore archive and do not delete.

That distinction is what this PR implements.

## Solution

A new `is_archived` flag on users. Archiving only hides a user. It never touches
`session_history` or `keep_history`, and it can be undone at any time.

Archived users are hidden from:

- the users list
- the history table (including the current activity union and the total duration)
- the home statistics and graphs
- the per library and per item user statistics
- the user selector on the history page

Both the users page and the history page get a **Show archived** toggle to bring them back
into view. Archived rows are dimmed and marked with an archive icon rather than simply
disappearing, so the state is visible rather than implicit.

Users can be archived from the users list in edit mode (a third inline toggle next to the
existing history and guest access toggles) or from the edit user dialog.

### Returning users

Users who regain access to the Plex server are unarchived automatically on the next user
refresh. Only users that were inactive *before* the refresh are considered, so a user that
was archived while still on the server stays archived. This is controlled by the new
**Automatically Unarchive Returning Users** setting (Settings > Plex Media Server > Server
Monitoring, advanced).

## Screenshots

**Users list, default and with "Show archived" enabled.** Archived users are dimmed and
marked with an archive badge rather than silently reappearing.

users_list_all

users_list_archived

**History, searching for an archived user.** The history is still there, it is just hidden
until archived users are shown.

history_hidden

history_show

**Archiving a user** from the users list in edit mode, or from the edit user dialog.

users_list_edit

user_settings

**The auto-unarchive setting.**

monitoring

## API

- new commands `archive_user` and `unarchive_user`
- `is_archived` accepted by `edit_user` and returned by `get_user`, `get_users` and
  `get_users_table`
- `include_archived` accepted by `get_history`, `get_user_list` and `get_user_names`

## Notes for review

- **Purely additive.** The migration adds one column with default `0`. With nothing
  archived, behaviour is identical to before.
- **Archived users are excluded from the aggregate statistics**, which is the one decision
  worth a second opinion. It is what makes archiving useful, since an archived user showing
  up under "Most Active Users" would defeat the point, but unlike the tables the totals
  change with no visible indication and no toggle next to them. The filtering is
  centralised, so it can be moved behind a setting or dropped if you would rather keep the
  aggregates untouched.
- `refresh_users()` reads the archived and inactive users *before* the upsert loop, because
  `plextv` returns `is_active = 1` for every user and the upsert would otherwise erase the
  transition that the auto-unarchive depends on.
- The archived filter in `get_datatables_history()` is appended *after* the session user
  filter, which matches on any clause containing `user_id`. Adding it earlier would fold
  the logged in guest's own id into the `NOT IN` list and hide their own history.

## Testing

Tested against a production database (55 users, ~17k history rows):

- archiving removes the user from the users list, the history table and the statistics
- the `session_history` row count for that user is unchanged
- unarchiving restores the previous state exactly
- verified on both a fresh install and an upgrade of an existing database

## Scope

This covers the user half of #1435. The issue also asks for the same treatment for movies,
shows and music that are no longer on the server, which is not part of this PR: there is no
per item equivalent of `deleted_section` to build on, so it needs a different approach and
would be better as its own change.

Refs #1435

## Type of Change

## Checklist


JonnyWong16:
I haven't reviewed this yet, but on first glance I think I like this implementation better than #2682.

PR #2780, opened 2026-08-10
Feature/storage capacity

JamesClayPatton:
## Description

Adds a Storage section to the homepage showing how much space is left on the drives your Plex libraries actually live on.

Tautulli already pulls the library list from Plex, and that response includes a `Location` for each library folder. This picks those up and works out which filesystem each one is on, so for the usual setup where Plex and Tautulli are on the same box there's nothing to configure.

Capacity comes from `shutil.disk_usage()`, one syscall per filesystem. Nothing walks the media folders, and the result is cached for 5 minutes. No new dependencies, and no extra requests to Plex since the folders come from the `/library/sections` response that's already being fetched.

Filesystems are keyed by device id rather than path, so if Movies, TV and Music all sit under `/mnt/media` you get one card instead of three. Bind mounts collapse the same way.

### Remote Plex and Docker

Plex reports paths as they exist on whatever machine Plex is running on, which Tautulli can't always see. Anything that can't be inspected gets grouped into a single "Unavailable" card rather than silently disappearing. Every filesystem call is wrapped, so a bad path can't take the homepage down with it.

There's an optional path mapping setting for when the paths don't line up (`/data/media|/mnt/media`). It's empty by default and you only need it if you're in that situation.

### Notes

The section and the API command are both admin only, since the cards show library folder paths. `get_storage_info` doesn't take a path argument, so there's no way to point it at arbitrary parts of the filesystem.

The section sits with the other homepage sections and toggles and reorders like them. I added it to the `HOME_SECTIONS` default, which only affects new installs since existing ones already have their list saved.

Tests are in `tests/test_storage.py` — plain `unittest`, nothing to install:

```
python -m unittest discover tests
```

They mock the filesystem calls and cover the dedup logic, Windows drive letters, UNC and POSIX paths, unreadable paths, the path mapping and the caching. A few run against the real filesystem.

### Scope

I know a broader version of this (disk space per disk on the Plex server) was looked at before and considered out of scope. This one is deliberately narrower: just the filesystems behind the library folders, only where Tautulli can actually see them, and opt-in. Happy to close it if it's still not something you want.

### Screenshot

The section itself:
01-storage-section

Sitting under the existing library stats:
02-in-context

In the homepage settings:
03-settings

And the path mapping field:
04-settings-mappings

## Type of Change

## Checklist


JonnyWong16:
I appreciate the effort, but monitoring disk usage is outside the scope of Tautulli. It's just not cross-platform compatible due to all the various ways Plex and Tautulli can be installed.

I just checked my analytics and over 70% of all Tautulli installs are using Docker. That means this feature would be useless for an overwhelming majority of users (they would just see an "Unavailable" card).

JamesClayPatton:
Makes sense, the Docker number is the part I didn't have. Thanks for taking the time to explain rather than just closing it.
 
I enjoyed working on this, and I'd like to keep contributing wherever needed. If there's anything else lying around that you'd want a hand with, I'm happy to take it on. 
Thanks!

Issue #2776, opened 2026-08-07
[Bug] Statistics Tiles Rendering Incorrectly


LRomandine:
### Describe the Bug

In Chrome 151 the statistics tiles have ugly static scrollbars on the right and bottom, and when scrolling in the tile the nice scrollbar appears (as they should).
<img width="516" height="190" alt="Image" src="https://github.com/user-attachments/assets/e7eaed05-e0e3-4e93-b332-ca94b7bae8b6" />

In Firefox 153.0.1 the numbers are cut off and I have to scroll right to view them (but the scrollbars are correct)
<img width="518" height="181" alt="Image" src="https://github.com/user-attachments/assets/e1e95e3a-3967-4eb3-9b08-aed8270c22c8" />

### Tautulli Version

2.17.2

### Git Branch

master

### Git Commit Hash

c7153deb6e3017b8dda46c8448fe548a61c5a47a

### Platform and Version

Ubuntu 24.04

### Python Version

3.13.13 (main, Apr 8 2026, 17:48:01) [GCC 14.2.0]

### Browser and Version

Chrome 151.0.7922.109

### Link to Logs

not needed

chase-roohms:
Thats so funny I was quite literally just coming to report the same bug, 

<img width="1061" height="753" alt="Image" src="https://github.com/user-attachments/assets/42f95cea-1449-4758-b3ae-2354ce062302" />

drlambnz:
Same issue here

<img width="1305" height="672" alt="Image" src="https://github.com/user-attachments/assets/e9006096-2570-44cd-89cd-338d173ed941" />

maldis018:
61b2e7e4 fixes the clipped counts, but two symptoms of this remain on Firefox 153.

The last row of each stats list is still unreachable. The plugin sets
`max-height: 132px` on `.scroll-content` for a 120px wrapper along with
`margin-bottom: -12px`, so 12px stays outside the wrapper's `overflow: hidden` even at
maximum scroll. Measured on nightly c6eceba9: all 15 stats tiles report
`scroll-content.offsetHeight - wrapper.clientHeight === 12`.

`.dashboard-activity-info-scroller` was not covered by the fix and still has the full
12px horizontal overhang, since the new rule is scoped to `.dashboard-stats-info-scroller`.

There is also a cosmetic side effect of the same overhang: the plugin sees the content as
12px wider than the wrapper and marks the x-axis scrollable, so a horizontal thumb renders
[...]

PR #2787, opened 2026-08-14
Fix remaining dashboard scroller clipping on Firefox 153

maldis018:
Follow-up to 61b2e7e4, which fixed the horizontal half of #2776. Two symptoms remain on Firefox 153.

`jquery.scrollbar` offsets `.scroll-content` by the measured native scrollbar size on
non-webkit browsers (`margin-right: -12px; margin-bottom: -12px`, plus
`max-height: 132px` for a 120px wrapper), expecting the native scrollbar to occupy that
overhang outside the wrapper's `overflow: hidden`. Firefox 153 honors the plugin's own
`::-webkit-scrollbar { width: 0; height: 0 }` as a hide-scrollbar shim, so the gutter is
0 and the overhang holds real content instead. Chrome never hit this because the plugin
short-circuits to `{width: 0, height: 0}` when `/webkit/i.test(navigator.userAgent)`.

`width: 100% !important` clamps the horizontal overhang, but `max-height: 132px` still
leaves 12px of vertical overhang, so the last row of a stats list is unreachable even at
maximum scroll. Measured on 2.17.2 and on nightly c6eceba9: all 15 stats tiles report
`scroll-content.offsetHeight - wrapper.clientHeight === 12`.

This adds `max-height: 100%` alongside `width: 100%`, and applies both to
`.dashboard-activity-info-scroller`, which has the same 12px horizontal overhang and was
not covered by 61b2e7e4. Percentages are safe on both because each wrapper has a fixed
pixel height (120px for stats, 225px for activity).

Verified on Firefox 153.0 and Chromium: zero horizontal and vertical overhang on all 16
scrollers, the phantom horizontal scrollbar is gone (the 12px overhang made the plugin
mark the x-axis scrollable while the native scroll range was 0, so the thumb rendered but
could not move), and vertical thumbs still appear on exactly the tiles with real overflow.

Related: #2762 targets the same clipping via row padding and `min-width` on the title
column, which addresses the layout symptom rather than the scroller overhang.

### Screenshots

latest (v2.17.2): counts column entirely clipped, last row cut
Screenshot_20260814_181041

nightly c6eceba9: counts restored by 61b2e7e4, last row still cut
Screenshot_20260814_182447

this PR: both fixed, last row fully visible at max scroll
Screenshot_20260814_180434

### Issues Fixed or Closed

- Refs #2776

## Type of Change

## Checklist

### DIFF
--- a/data/interfaces/default/css/tautulli.css
+++ b/data/interfaces/default/css/tautulli.css
@@ -997,6 +997,10 @@ a .users-poster-face:hover {
     flex-grow: 1;
     z-index: 1;
 }
+.dashboard-activity-info-scroller.scroll-wrapper > .scroll-content {
+    width: 100% !important;
+    max-height: 100% !important;
+}
 .dashboard-activity-info-scroller.scrollbar-macosx > .scroll-element.scroll-y {
     left: 10px;
 }
@@ -1442,6 +1446,7 @@ a .dashboard-activity-metadata-user-thumb:hover {
 }
 .dashboard-stats-info-scroller.scroll-wrapper > .scroll-content {
     width: 100% !important;
+    max-height: 100% !important;
 }
 .dashboard-stats-info-scroller.scrollbar-macosx > .scroll-element.scroll-y {
     left: 10px;


JonnyWong16 (approved):
Thanks. I didn't know about the vertical cutoff.

PR #2800, opened 2026-08-23
Change PMS Image Proxy to Explicit

aauren:
## Description

Tautulli's session tool defaults to 'early' locking, which holds a per-session lock across the entire request, and /pms_image_proxy is by far the slowest handler on the dashboard since it round-trips to Plex for a transcode. With a shared browser session this serializes every image request behind whichever one is already in flight, even though the browser opens six concurrent connections.

Overriding locking to 'explicit' for this endpoint alone removes that serialization, since the handler never reads or writes session data and has nothing to protect. Measured end to end this took a cold dashboard load from roughly 5.4s to 2s without changing session behavior anywhere else in the app.

### Screenshot

N/A

### Issues Fixed or Closed

N/A (no issue currently reported)

## Type of Change

- [ ] New feature (non-breaking change which adds functionality)

Improves performance of Tautulli by not locking on image proxy requests.

## Checklist

- [ ] I have added or updated the docstring for new or existing methods

### DIFF
--- a/plexpy/webstart.py
+++ b/plexpy/webstart.py
@@ -253,6 +253,10 @@ def initialize(options):
             'tools.auth_basic.on': False,
             'tools.sessions.on': False,
         },
+        '/pms_image_proxy': {
+            # Slowest handler (Plex round-trip) but GET-only and never touches cherrypy.session, so nothing to lock.
+            'tools.sessions.locking': 'explicit',
+        },
         '/status': {
             'tools.auth_basic.on': False,
             'tools.sessions.on': True,


JonnyWong16:
@claude review

JonnyWong16:
Thanks for the PR, but session locking is already set to `'explicit'` globally so this change isn't needed.

aauren:
Ahh... I didn't notice this come in: https://github.com/Tautulli/Tautulli/commit/a1b3cb8e4285d06b513fa20f1787113ac756ca36

I had been carrying this change for a bit locally before I upstreamed it.

Thanks for the consideration anyway, and sorry for the noise!

